PR #30, opened 2020-09-24
CoverPage change for large numbers of messages

jgibbon:
The CoverPage wasn't quite prepared for users residing in really spammy channels.

Changes:
 - As suggested, getShortenedCount was moved from ChatPage to functions.js and is now also used in the CoverPage
 - A small copy/paste error was fixed in the CoverPage

Before/after:
[](https://i.imgur.com/YDn8WnW.png)

As always, it's a pleasure to "work" with you,
cheers!

John


Wunderfitz:
Thank you very much!

PR #35, opened 2020-09-27
Build requires Qt5DBus

monich:
I also suggest to drop the .yaml file. AFAICT it's useless.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -104,3 +104,5 @@ HEADERS += \
     src/tdlibsecrets.h \
     src/tdlibwrapper.h
 
+OTHER_FILES += \
+    rpm/$${TARGET}.spec
--- a/rpm/harbour-fernschreiber.spec
+++ b/rpm/harbour-fernschreiber.spec
@@ -23,6 +23,7 @@ BuildRequires:  pkgconfig(sailfishapp) >= 1.0.2
 BuildRequires:  pkgconfig(Qt5Core)
 BuildRequires:  pkgconfig(Qt5Qml)
 BuildRequires:  pkgconfig(Qt5Quick)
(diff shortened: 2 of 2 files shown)


harbour-fernschreiber.pro
```diff
@@ -104,3 +104,5 @@ HEADERS += \
     src/tdlibsecrets.h \
     src/tdlibwrapper.h
 
+OTHER_FILES += \
+    rpm/$${TARGET}.spec
```
Wunderfitz, line 108: Any particular reason why to add OTHER_FILES and only to add the spec file here?

Wunderfitz (commented):
Thanks for adding the DBus build dependency. However, please clarify why OTHER_FILES and only the spec is needed...

harbour-fernschreiber.pro
monich: If I started adding other files, then I would have to add all other files too, including the .yaml thing which I think is useless. Adding two out of three files would look strange, and what if you want to keep the .yaml file? So I just added the file which I edited ))

Wunderfitz: But why is the OTHER_FILES section needed at all? All elements are listed in DISTFILES, so what's the point of OTHER_FILES?

monich: Just for convenience of editing, I don't think it has any other meaning

Wunderfitz: :D Well, in this case I might consider deleting this one again. Let's see!

monich: Well, it's up to you. Spec isn't being edited that often. I personally prefer to have all files included into the project and accessible directly from QtCreator but it's a matter if taste and habits of course.

PR #37, opened 2020-09-27
Fix chat ordering and optimize chat data storage

monich:
This makes the chat order look reasonable to me.

And I think I figured how to hide irrelevant chats but that would be a separate PR, if and after this one gets merged.

### DIFF
--- a/src/tdlibreceiver.cpp
+++ b/src/tdlibreceiver.cpp
@@ -18,25 +18,93 @@
 */
 #include "tdlibreceiver.h"
 
+#define LOG(x) qDebug() << "[TDLibReceiver]" << x
+
+namespace {
+    const QString ID("id");
+    const QString LIST("list");
+    const QString CHAT_ID("chat_id");
+    const QString POSITION("position");
+    const QString POSITIONS("positions");
+    const QString ORDER("order");
+    const QString LAST_MESSAGE("last_message");
+    const QString UNREAD_COUNT("unread_count");
+    const QString LAST_READ_INBOX_MESSAGE_ID("last_read_inbox_message_id");
+    const QString LAST_READ_OUTBOX_MESSAGE_ID("last_read_outbox_message_id");
+
+    const QString TYPE("@type");
+    const QString TYPE_CHAT_POSITION("chatPosition");
+    const QString TYPE_CHAT_LIST_MAIN("chatListMain");
+}
+
+static QString getChatPositionOrder(const QVariantMap &position)
+{
+    if (position.value(TYPE).toString() == TYPE_CHAT_POSITION &&
+        position.value(LIST).toMap().value(TYPE) == TYPE_CHAT_LIST_MAIN) {
+        return position.value(ORDER).toString();
+    }
+    return QString();
+}
+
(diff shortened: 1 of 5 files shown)


monich:
For the record, I'm building and testing against [this](https://github.com/monich/td/tree/rpm) version of tdlib (which might be different from yours).

Wunderfitz:
Mhmm, quite a lot to review - that might take a while. Moreover, I need to make up my mind if I accept the code style changes... ;)

In the meantime: Could you maybe explain a bit the reasoning behind the changes and which bugs you fixed?

monich:
Overall it's more or less described in the commit messages. It's about chat ordering (handling `"updateChatPosition"`, extracting chat position from `"updateChatLastMessage"`, maintaining the model order without re-sorting the whole thing) and performance (to avoid "App is not responding, close/wait?" popups and generally improve the responsiveness).

If you're puzzled with particular changes, ask. To me, most of them seem to be fairly obvious.

I have a few more ideas about optimization (storing more numbers as numbers rather than strings/variants, exposing more roles from the model to avoid re-drawing things that haven't changed, etc.) but let's see if you accept any code changes at all, before I dive deeper into that.

And as I already mentioned, I have a plan how to fix the problem with rogue chats littering the chat list, that also depends on whether or not this one gets merged.

src/tdlibreceiver.cpp
```diff
@@ -164,123 +212,139 @@ void TDLibReceiver::processUpdateUnreadChatCount(const QVariantMap &receivedInfo
...
     emit unreadChatCountUpdated(chatCountInformation);
 }
 
 void TDLibReceiver::processUpdateChatLastMessage(const QVariantMap &receivedInformation)
 {
-    QVariantMap lastMessage = receivedInformation.value("last_message").toMap();
-    qDebug() << "[TDLibReceiver] Last message of chat " << receivedInformation.value("chat_id").toString() << " updated, order " << receivedInformation.value("order").toString() << " content: " << lastMessage.value("content").toString();
-    emit chatLastMessageUpdated(receivedInformation.value("chat_id").toString(), receivedInformation.value("order").toString(), lastMessage);
+    const QString chat_id(receivedInformation.value(CHAT_ID).toString());
+    const QString order(findChatPositionOrder(receivedInformation.value(POSITIONS).toList()));
```
Wunderfitz, line 222: This doesn't work with TDLib 1.6.0 - the change with 'positions' and 'updateChatPosition' was introduced in May with https://github.com/tdlib/td/commit/06136146511a46bcc26fcb936e3e9ae42130e92c, so after 1.6.0. However, a small if/else seems to do the trick - so no blocker. :)

Wunderfitz (approved):
Thanks again for optimizing the sorting and all the other things. Seems to work properly with a small exception (probably as you are using the bleeding-edge state and I the released version 1.6.0). Let's merge it and continue from there. :+1:

src/tdlibreceiver.cpp
monich: Good to know, I'll switch my build to 1.6.0 too.

Issue #36, opened 2020-09-27
Display shared Locations


jgibbon:
Hi,
I'd like to take a look at adding this, but obviously don't want to implement stuff you wouldn't like.
Please let me know what you think about these points or if you'd even like me to have a go at this to get my feet wet with Fernschreiber:

### General
There are only a few ways to implement showing maps, all with pros and cons.
 - Static map images: While not very resource intensive, the only online service I found that doesn't require an API key is yandex. It only supports small images. But even with other services, I see possible privacy concerns.
- QT Location: Pretty resource intensive (perhaps too much so to integrate into the chat list), also obviously not harbour compliant. For the detail view, it may still the best option **if you do not plan to submit fernschreiber to the Jolla Store**. There could be a possibility to set user-defined OSM-Tile-URLs in the options, too. (If you like, I could also adapt the settings ui from Talefish in case there'll be a lot of options at some point – but that's another topic)

### Chat List (LocationPreview)
 - While static map images would work (it's an image, after all), it may not be desirable for users to send location data to yandex/google/…
 - As mentioned above, displaying a "real" map would be pretty but _could_ get a bit problematic performance-wise.
 - Just displaying a Button (to get to the LocationPage) with the adress doesn't work – reverse geocoding seems to be broken with osm and/or the available QT Location library. Displaying the coordinates or even "Show Location" instead would be fast/easy, but a bit ugly.
 - Another option would be a Button by default with an option to use a (yandex?) map image in the application settings.

### Detail view (LocationPage)
 - Using a Map from QT Location is relatively straight forward. A slight problem is accidentally going back in the PageStack while swiping around. This can, to a degree, be mitigated by disabling "back navigation" on the page for a while when the map center is changed.
 - I don't know yet how to handle message updates for live locations, haven't had a look this deep.
 - While it would be nice to have all currently shared live locations of a chat displayed on one map, I think it's okay not to do that at least initially.

Feel free to get in touch via Telegram, as well, to discuss things if you want to.

cheers,
"velox"

Wunderfitz:
[...]

PR #38, opened 2020-09-28
Handle display of messageLocation

jgibbon:
Hi @Wunderfitz,
I have implemented the rudimentary display of the messageLocation type pretty much as planned in #36. 

What this does:
 - LocationPreview is almost exactly like the existing ImagePreview, except it first requests the map image file from Telegram.
 - When the image is clicked, the new ProcessLauncher class tries to start Pure Maps. If that fails, a Notification is shown.
- ~~This does not handle live location updates yet, so closing the issue should wait.~~ Nonetheless I found it useful enough to warrant a quick PR.

A small drawback is that I've found no way to really map the response of getMapThumbnailFile to the actual chat message. This implementation simply guesses that the first onFileUpdated response should be the matching one. While this worked well in my tests, it leaves a bit of an aftertaste.

Cheers,
John/velox

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -24,6 +24,7 @@ SOURCES += src/harbour-fernschreiber.cpp \
     src/dbusadaptor.cpp \
     src/dbusinterface.cpp \
     src/notificationmanager.cpp \
+    src/processlauncher.cpp \
     src/tdlibreceiver.cpp \
     src/tdlibwrapper.cpp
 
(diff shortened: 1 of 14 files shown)


jgibbon:
Had a quick look at live locations in my lunch break. 
Do you think that the update should be triggered from within the Preview instead of the ChatPage to minimize things to run for other edits?

Wunderfitz:
LGTM, thank you very much! Smaller open details can be ironed out later. ;)

PR #40, opened 2020-09-30
Expose more roles from ChatListModel

monich:
Roles can now be directly bound to individual UI elements.

Replaced per-delegate timestamp refresh timers with a single one refreshing all timestamps with a single `dataChanged` signal.

Removed `chatChanged` signal since it's no longer necessary.

This should fix a few small UI issues (periodic flicking, strange behavior after long-tapping your own chat), but it's mostly about optimization and deletion of duplicate QML code.

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -231,8 +231,8 @@ Page {
                             pageStack.push(Qt.resolvedUrl("../pages/ChatPage.qml"), { "chatInformation" : display });
                         }
 
+                        showMenuOnPressAndHold: chat_id != overviewPage.ownUserId
                         menu: ContextMenu {
-                            visible: display.id !== overviewPage.ownUserId
                             MenuItem {
                                 onClicked: {
                                     var newNotificationSettings = display.notification_settings;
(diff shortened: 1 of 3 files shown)


Wunderfitz:
Thank you very much - I'll review (and most certainly) merge on the weekend. Need some break now. ;)

monich:
Sure, take your time, I have other work to do as well!

qml/pages/OverviewPage.qml
```diff
@@ -349,7 +333,7 @@ Page {
                                         spacing: Theme.paddingSmall
                                         Text {
                                             id: chatListLastUserText
-                                            text: (typeof display.last_message !== "undefined") ? ( display.last_message.sender_user_id !== overviewPage.ownUserId ? Emoji.emojify(Functions.getUserName(tdLibWrapper.getUserInformation(display.last_message.sender_user_id)), font.pixelSize) : qsTr("You") ) : qsTr("Unknown")
+                                            text: last_message_sender_id ? ( last_message_sender_id !== overviewPage.ownUserId ? Emoji.emojify(Functions.getUserName(tdLibWrapper.getUserInformation(last_message_sender_id)), font.pixelSize) : qsTr("You") ) : qsTr("Unknown")
```
Wunderfitz, line 336: This seems to change the behavior for channels implicitly. Now it displays "Unknown" instead of an empty string. Not a real blocker, in a later change I will make the text invisible if it's belonging to a channel....

Wunderfitz (approved):
Thank you very much, LGTM. :)

qml/pages/OverviewPage.qml
monich: Right, good catch. Too many "Unknown"s just litters the display. I wonder if this "Unknown" was needed at all.

Wunderfitz: It was a failsafe string from the very beginning. However, it seems that it's only really needed for conversations with deleted users.

PR #42, opened 2020-10-01
handle user registration

jgibbon:
First of all: 
Sorry, most parts from this one are a bit tricky to test. I've only been able to go through the sign up flow (once!), because I had a spare SIM laying around that never had anything to do with Telegram. Also, I apologize for including unrelated changes – those things bugged me a bit.

Relevant changes:
- WaitRegistration authorizationState is now handled. 
- If it is encountered, the user is asked to enter a first name and a last name. 
- If Telegram demands it, a server-supplied "terms of service" text with an "OK" Button is shown before that.
- When the authorization is done, there is a new check for an empty "chats", object. Previously, the other signals listened to were never emitted, so OverviewPage was stuck in Loading when there were no chats.
- If there are no Chats, a ViewPlaceholder is shown. Granted, there's not much else to do for the user after that.
- OverviewPage: All Columns now reside in the Flickable. Please check if the resulting layout changes are OK with you.

Unrelated changes (sorry):
- There was an error in CoverPage that annoyed me all the time, so I just put it in here, as well.
- After rebasing to your upstream, I found the architecture was changed to i486. Obviously I had to prevent an architecture ping pong from happening again. I did not have the chance to test the i486 build because I don't have TDlib for that yet. But it complains and fails, which is a good sign.

cheers,
John

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ const char TDLIB_API_HASH[] = "1234567890abcdef1234567890abcdef";
 
 You get the Telegram API ID and hash as soon as you've registered your own application on [https://my.telegram.org](https://my.telegram.org).
 
-Moreover, you need to have a compiled version of [TDLib](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`. Before compiling the project, be sure to set the target architecture in the project file `harbour-fernschreiber.pro` correctly using the variable `TARGET_ARCHITECTURE`. This can most certainly be automated. Pull requests welcome! ;)
+Moreover, you need to have a compiled version of [TDLib](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
 
 ## Credits
 This project uses
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -147,6 +147,12 @@ Page {
                 chatListCreatedTimer.start();
             }
         }
+        onChatsReceived: {
+            if(chats && chats.chat_ids && chats.chat_ids.length === 0) {
+                chatListCreatedTimer.stop();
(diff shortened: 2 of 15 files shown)


qml/pages/OverviewPage.qml
```diff
@@ -152,6 +147,12 @@ Page {
                 chatListCreatedTimer.start();
             }
         }
+        onChatsReceived: {
+            if(chats && chats.chat_ids && chats.chat_ids.length === 0) {
+                chatListCreatedTimer.stop();
+                chatListCreatedTimer.start();
```
monich, line 153: These two can be replaced with a single `restart()`

src/tdlibwrapper.cpp
```diff
@@ -709,6 +725,11 @@ void TDLibWrapper::handleMessagesDeleted(const QString &chatId, const QVariantLi
     emit messagesDeleted(chatId, messageIds);
 }
 
+void TDLibWrapper::handleChats(const QVariantMap &chats)
+{
+    emit this->chatsReceived(chats);
```
monich, line 730: I believe you can directly connect two signals like this:
```c++
connect(this->tdLibReceiver, SIGNAL(chats(QVariantMap)), this, SIGNAL(chatsReceived(QVariantMap)));
```

README.md
```diff
@@ -34,7 +34,7 @@ const char TDLIB_API_HASH[] = "1234567890abcdef1234567890abcdef";
...
 
-Moreover, you need to have a compiled version of [TDLib](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`. Before compiling the project, be sure to set the target architecture in the project file `harbour-fernschreiber.pro` correctly using the variable `TARGET_ARCHITECTURE`. This can most certainly be automated. Pull requests welcome! ;)
+Moreover, you need to have a compiled version of [TDLib](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
```
monich, line 37: Hmm, pull requests are not welcome anymore :|

Wunderfitz (approved):
Worked smoothly with the test DC (can be set with use_test_dc = true in TDLibWrapper::setInitialParameters()), also the other changes make sense, so let's merge them. If there are any regressions/bugs, they can be fixed later. ;)

Wunderfitz:
Thanks again! :)

PR #44, opened 2020-10-03
Enable verbose output in TDLibReceiver if DEBUG is defined

monich:
This should have no effect on release build but helps me with debugging.


Wunderfitz:
Thx again! :)

Issue #2, opened 2020-09-20
Chat Overview: Items not correctly updated sometimes...


Wunderfitz:
In certain cases, some items of the chat overview are not correctly updated if there are new messages. Can be mitigated by scrolling down and up to redraw everything (or by a restart of the app), but probably we can do better...

Wunderfitz:
Seems to be fixed by several changes, big thanks especially to @monich

PR #45, opened 2020-10-04
Hide irrelevant chats

monich:
`chatListModel.showAllChats` property can be used to switch visibility of irrelevant groups on and off. There's no UI for it though, and I'm not 100% sure if it's a good idea to have one :| It's not something that people are likely to use often.

Basically, the criteria for including chats into the list is the chat member status. Those with `chatMemberStatusLeft` are left out of the model by default. Tested with both basic groups and supergroups.

It's a pretty big one, must be a pain to review. But you asked for contributions, here we go))

### DIFF
--- a/src/chatlistmodel.cpp
+++ b/src/chatlistmodel.cpp
@@ -27,6 +27,7 @@ namespace {
     const QString ID("id");
     const QString DATE("date");
     const QString TEXT("text");
+    const QString TYPE("type");
     const QString TITLE("title");
     const QString PHOTO("photo");
     const QString SMALL("small");
@@ -35,6 +36,8 @@ namespace {
     const QString CONTENT("content");
     const QString LAST_MESSAGE("last_message");
     const QString SENDER_USER_ID("sender_user_id");
+    const QString BASIC_GROUP_ID("basic_group_id");
+    const QString SUPERGROUP_ID("supergroup_id");
     const QString UNREAD_COUNT("unread_count");
     const QString NOTIFICATION_SETTINGS("notification_settings");
     const QString LAST_READ_INBOX_MESSAGE_ID("last_read_inbox_message_id");
@@ -42,8 +45,12 @@ namespace {
     const QString TYPE_MAP("type");
     const QString IS_CHANNEL("is_channel");
 
-    const QString TYPE("@type");
+    const QString _TYPE("@type");
     const QString TYPE_MESSAGE_TEXT("messageText");
+    const QString TYPE_CHAT_TYPE_PRIVATE("chatTypePrivate");
+    const QString TYPE_CHAT_TYPE_BASIC_GROUP("chatTypeBasicGroup");
+    const QString TYPE_CHAT_TYPE_SUPERGROUP("chatTypeSupergroup");
+    const QString TYPE_CHAT_TYPE_SECRET("chatTypeSecret");
 }
 
 class ChatListModel::ChatData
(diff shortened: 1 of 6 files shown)


monich:
I think I better add a few comments explaining some of the changes which may seem strange from the first glance.

1. Whenever you do `map.value("foo")`, a temporary `QString` is being constructed on stack, which takes ASCII string as a parameter, converts it into internal UTF-16 representation and that temporary `QString` object is used as a key for doing the lookup in `QVariantMap`. That's a runtime overhead. A pre-allocated static `const QString` allows to do the initialization and ASCII => UTF-16 conversion once in the lifetime of the application.
2. Another way to avoid the above mentioned overhead is `QStringLiteral`. Even though it's even more efficient than static `QString`, I don't really like those - IMO it litters the code too much. But that's a matter of taste of course.
3. Comparing primitive types (ints and such) and using them as keys in a hashtable is significantly more efficient than using `QString` representation of the number for the same purpose, both in terms and memory usage and CPU utilization. Object ids are coming in JSON format as numbers (without quotes), converting them to `QString` and dragging `QString` around is a waste of resources.
4. Storing data in `QVariantMap` is pretty convenient but if lookups and especially conversions are being done often enough, it may be a good idea to extract critical pieces of information (such as object ids) out of `QVariantMap` and store them separately in a ready to use form. In addition to being more efficient at run time, it provides better type checking at compile time. By treating everything as a `QVariant` you shift all the checking to run time.

If something else looks funny or suspicious, go ahead and ask!

Wunderfitz:
Thanks for the great explanation - as mentioned, it's still a learning exercise for me. I Hope I can review it in the course of the day, but @jgibbon's PR comes first. I hope that there are no conflicts. 

Moreover, it seems that the last change re-introduced an issue I had a long time in the chat list - the text of the last message in a chat isn't always updated properly - that was one reason for the QML-side update timer. I haven't seen issues for other elements like the profile pictures or the last user name, but we'll see... Maybe I want to hunt down this nasty bug first before I merge other things...

monich:
What do you mean "the last message in a chat isn't always updated properly"?

monich:
Never mind, just noticed https://github.com/Wunderfitz/harbour-fernschreiber/issues/2

Wunderfitz:
Yep and I had the impression that with 0.2 it was gone. Yes, there was this flickering because of the update, but at least everything was consistent. I could reproduce it now several times this morning with the latest change, so from my perspective we might need some additional QML-based update there to ensure that everything is fine eventually. Makes me sad as well...

src/chatlistmodel.cpp
```diff
@@ -35,15 +36,21 @@ namespace {
...
+    const QString SUPERGROUP_ID("supergroup_id");
     const QString UNREAD_COUNT("unread_count");
     const QString NOTIFICATION_SETTINGS("notification_settings");
     const QString LAST_READ_INBOX_MESSAGE_ID("last_read_inbox_message_id");
     const QString LAST_READ_OUTBOX_MESSAGE_ID("last_read_outbox_message_id");
     const QString TYPE_MAP("type");
     const QString IS_CHANNEL("is_channel");
 
-    const QString TYPE("@type");
+    const QString _TYPE("@type");
```
Wunderfitz, line 48: Well, there was a TYPE_MAP constant in there, but I can adjust that later as well... ;)

Wunderfitz (approved):
Seems to work - at least in my first tests. :) 

I'm pretty much aware of the overhead when encoding primitive types with Strings, but I had terrible issues at the beginning on the QML/JS side with the big integers, so ended up using Strings. 

Seems to work now, so let's see if we have some regressions or not. No matter what it's still a 0.something, so we can still deal with them if they come up, I suppose.

PR #51, opened 2020-10-04
OBS build requires qt5-qttools-linguist

monich:
Otherwise translations will be missing. That's not the only thing that breaks OBS build though (most important reasons - dependency on tdlib and tdlibsecrets.h) so it doesn't really matter... It's up to you. Those who want to do reproducible unattended builds would have to modify .spec file anyway, this would just reduce the amount of changes they have to make.


Wunderfitz:
Well, if it makes life a little easier, let's merge it. Concerning the other dependencies: I'm also open to changes there, e.g. if API key and hash can be injected via variables that might also be beneficial for automated builds.

PR #55, opened 2020-10-05
Update harbour-fernschreiber-es.ts

GNUuser:
What is "Avenue" exactly?

### DIFF
--- a/translations/harbour-fernschreiber-es.ts
+++ b/translations/harbour-fernschreiber-es.ts
@@ -13,7 +13,7 @@
     </message>
     <message>
         <source>Send E-Mail</source>
-        <translation>Enviar C.Electron.</translation>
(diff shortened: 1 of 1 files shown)


Wunderfitz:
"Venue" is probably hard to translate in many languages - we don't have a proper expression in German at least. If I recall my Spanish from school and university correctly, "lugar" is a pretty good choice. In the end it's a place with a given address - coming from Foursquare.

Muchas gracias por la actualisación...

GNUuser:
I think "Venue" is  a sentence to use it in GPS/MAPS apps, because "venue" is same with "Street" i just saying...

Wunderfitz:
No, venue is a certain place, see for example: https://www.merriam-webster.com/dictionary/venue. "Avenue" would be a synonym to "street"...

Issue #23, opened 2020-09-21
configurable UI


chstem:
Here are a couple of suggestions that may make the UI more compact. Since a "good" UI is strongly based on user preferences, all points could be implemented such that they are configurable (either individually or by a single "compact UI" switch).
 
### more compact chat overview page
- [ ] removing the horizontal line separator
- [ ] smaller chat icons
- [ ] showing less information about each chat

### more simplistic chat page
- [ ] option to remove/hide the frames around each message (e.g. depecher allows to change opacity from 0 to 1)
- [ ] format the users own text left-justified (but keep the right-alignment of the overall box)
- [ ] do not highlight unread messaged (if you scroll down through the chat history, they effectively just blink)
- [ ] as a possible alternative: insert a single "unread messages" marker (similar to depecher)

ferlanero:
Bubbles would have to have an option to disable. They overcharge viewing and adds an unnecessary complexity to chats view as I already have commented on #223

PR #56, opened 2020-10-06
Moved app settings to a separate object

monich:
As more settings seem to be coming (https://github.com/Wunderfitz/harbour-fernschreiber/pull/54, https://github.com/Wunderfitz/harbour-fernschreiber/issues/23) I think it's a good idea to separate them into a separate object. `TDLibWrapper` is already pretty big and is going to grow even bigger.

This object exposes settings (well, currently just one) as properties as opposed to invokable getters/setters. Properties are good for bindings, automatic re-evaluation of expressions, their values get cached by QML runtime and they generally fit nicely into QML philosophy.

This change doesn't affect any functionality yet, it's just a groundwork for the upcoming changes.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -19,6 +19,7 @@ PKGCONFIG += nemonotifications-qt5 ngf-qt5
 QT += core dbus
 
 SOURCES += src/harbour-fernschreiber.cpp \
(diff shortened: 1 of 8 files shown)


Wunderfitz:
A very good idea, thank you very much! :+1:

PR #58, opened 2020-10-06
Allow to delete messages from your own private chat

monich:
They have `"can_be_deleted_for_all_users": false` and yet you can delete those because you are the only one who has access to your own private chat.


Wunderfitz:
Good point - thanks for correcting that!

PR #60, opened 2020-10-09
Made "Stickers as images" feature optional

monich:
Since you're saying that optimization efforts have to be consistently applied to the entire files or else they are not being accepted, I did my best to stop myself from doing any optimization this time. Expressions `display.content['@type'] === "messageSticker"`and `chatPage.myUserId === display.sender_user_id` are being (consistently!) evaluated as many times as possible.

### DIFF
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -497,7 +497,6 @@ Page {
                             onTriggered: {
                                 webPagePreviewLoader.active = ( typeof display.content.web_page !== "undefined" );
                                 imagePreviewLoader.active = ( display.content['@type'] === "messagePhoto" );
-                                stickerPreviewLoader.active = ( display.content['@type'] === "messageSticker" );
                                 videoPreviewLoader.active = (( display.content['@type'] === "messageVideo" ) || ( display.content['@type'] === "messageAnimation" ));
                                 audioPreviewLoader.active = (( display.content['@type'] === "messageVoiceNote" ) || ( display.content['@type'] === "messageAudio" ));
                                 documentPreviewLoader.active = ( display.content['@type'] === "messageDocument" );
@@ -560,6 +559,7 @@ Page {
                                     color: index > ( chatView.count - chatInformation.unread_count - 1 ) ? Theme.secondaryHighlightColor : Theme.secondaryColor
                                     radius: parent.width / 50
                                     opacity: index > ( chatView.count - chatInformation.unread_count - 1 ) ? 0.5 : 0.2
+                                    visible: appSettings.showStickersAsImages || display.content['@type'] !== "messageSticker"
                                 }
 
                                 Column {
@@ -740,18 +740,16 @@ Page {
                                     Component {
                                         id: stickerPreviewComponent
                                         StickerPreview {
-                                            id: messageStickerPreview
                                             stickerData: ( display.content['@type'] === "messageSticker" ) ?  display.content.sticker : ""
-                                            visible: display.content['@type'] === "messageSticker"
-                                            anchors.horizontalCenter: parent.horizontalCenter
                                         }
                                     }
 
                                     Loader {
                                         id: stickerPreviewLoader
-                                        active: false
+                                        active: display.content['@type'] === "messageSticker"
                                         asynchronous: true
-                                        width: parent.width
+                                        x: (chatPage.myUserId === display.sender_user_id) ? (parent.width - width) : 0
+                                        width: (appSettings.showStickersAsImages || !item) ? parent.width : item.width
                                         sourceComponent: stickerPreviewComponent
                                     }
 
--- a/qml/pages/SettingsPage.qml
+++ b/qml/pages/SettingsPage.qml
@@ -33,7 +33,6 @@ Page {
         Column {
             id: column
             width: settingsPage.width
-            spacing: Theme.paddingLarge
 
             PageHeader {
                 title: qsTr("Settings")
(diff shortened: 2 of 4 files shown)


qml/pages/ChatPage.qml
```diff
@@ -497,7 +497,6 @@ Page {
                             onTriggered: {
                                 webPagePreviewLoader.active = ( typeof display.content.web_page !== "undefined" );
                                 imagePreviewLoader.active = ( display.content['@type'] === "messagePhoto" );
-                                stickerPreviewLoader.active = ( display.content['@type'] === "messageSticker" );
```
monich, line 500: Sorry, a bit of an inconsistency is introduced here, but I really want stickers to be displayed right away rather than after 500ms timeout.

qml/pages/SettingsPage.qml
```diff
@@ -46,15 +45,29 @@ Page {
             TextSwitch {
                 checked: appSettings.sendByEnter
                 text: qsTr("Send message by enter")
-                description: qsTr("Send your message by pressing the enter key")
+                description: qsTr("Send your message by pressing the enter key.")
```
Wunderfitz, line 48: If you do that, it invalidates all translations for this text. So, if you want to introduce periods at the end, please also change all existing translations to avoid follow-up work by others.

qml/pages/SettingsPage.qml
```diff
@@ -46,15 +45,29 @@ Page {
...
 
-            VerticalScrollDecorator {}
+            SectionHeader {
+                text: qsTr("Appearance")
+            }
+
+            TextSwitch {
+                checked: appSettings.showStickersAsImages
+                text: qsTr("Show stickers as images")
+                description: qsTr("Normally stickers don't have a background and aligned like text, but you can change that and display them like regilar images.")
```
Wunderfitz, line 62: Please don't put valuations ("normally") in descriptive texts and check your spelling ("regilar"). Better: "Show background for stickers and align them centrally like images"

src/appsettings.cpp
```diff
@@ -41,3 +42,17 @@ void AppSettings::setSendByEnter(bool sendByEnter)
         emit sendByEnterChanged();
     }
 }
+
+bool AppSettings::showStickersAsImages() const
+{
+    return settings.value(KEY_SHOW_STICKERS_AS_IMAGES, false).toBool();
```
Wunderfitz, line 48: Default should be true.

Wunderfitz (changes requested):
Thanks for your changes. I have some remarks though. If they were minor, I'd have done the necessary changes on my own, but it seems that they would cause a little more (especially with the existing translations) if we left it as is.

qml/pages/SettingsPage.qml
monich: You can drop it, it's just that most text switch descriptions in Sailfish OS (although not all, so it's not 100% consistent even there!) are full sentences terminated with a period. Sounds like it's not worth the trouble.

monich: OK, I'll drop it.

qml/pages/SettingsPage.qml
monich: Done.

src/appsettings.cpp
monich: Hmm... but you said you don't want to surprise people!

monich: Oh well, it's your code. If you want to surprise, let it be a surprise.

qml/pages/ChatPage.qml
```diff
@@ -740,18 +740,16 @@ Page {
...
                                             stickerData: ( display.content['@type'] === "messageSticker" ) ?  display.content.sticker : ""
-                                            visible: display.content['@type'] === "messageSticker"
-                                            anchors.horizontalCenter: parent.horizontalCenter
                                         }
                                     }
 
                                     Loader {
                                         id: stickerPreviewLoader
-                                        active: false
+                                        active: display.content['@type'] === "messageSticker"
```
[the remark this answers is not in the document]
monich: I suspect that this delay is responsible for annoying jumping and flicking during regular scrolling - delegates are initialized with wrong height and then get resized after a delay, it looks pretty chaotic, especially with images. I have a feeling that asynchronous loading and a bit of optimization could make both slow and fast scroll work reasonably well without this delay.

qml/pages/ChatPage.qml
```diff
@@ -740,18 +740,16 @@ Page {
...
                                         asynchronous: true
-                                        width: parent.width
+                                        x: (chatPage.myUserId === display.sender_user_id) ? (parent.width - width) : 0
```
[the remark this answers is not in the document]
monich: The alignment of sticker inside of the delegate is taken care by the next line (which calculates the width of the sticker preview component).

I don't think you can use bindings to adjust anchors. You would have to go through defining states and associated  [AnchorChanges](https://doc.qt.io/qt-5/qml-qtquick-anchorchanges.html) which is just too much trouble for something as simple as this.

qml/pages/ChatPage.qml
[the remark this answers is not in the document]
Wunderfitz: Well, the jumping is also there if you load it immediately. To avoid that you probably need to reserve the respective space upfront, but I haven't digged deeper there yet. I also have the same issue with Piepmatz BTW.

Wunderfitz (approved):
Thx for the changes, LGTM.

qml/pages/ChatPage.qml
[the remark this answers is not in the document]
monich: AFAICT the size of the image is known in advance, meaning that delegate can be sized right away, even before the actual content has been loaded.

monich:
Yey, thanks for merging! Now it looks like real Telegram to me))

In some chats where I spent my time, people sometimes literally "talk" with stickers, i.e. at times there are more stickers than text on the screen. Displaying them right it an important part of the user experience!

qml/pages/ChatPage.qml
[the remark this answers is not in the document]
Wunderfitz: To a certain extent the height is known, yes. There is some uncertainty involved, but any precalculation is probably better than the large resizing procedures later. I introduced some changes in line with your suggestions on the Loader/Component handling with 67669cf63f15e04452cf51c3edfe9bc4f29c8ec5 and d3295bc49ac0276971507837e7410eebbedbbefb. Let's see how it performs...

PR #61, opened 2020-10-10
Toggle "Image only" mode on tap

monich:
Sometimes I want to look at the image without anything else being painted on top of it. I'm sure many others would appreciate that too.

### DIFF
--- a/qml/pages/ImagePage.qml
+++ b/qml/pages/ImagePage.qml
@@ -25,6 +25,7 @@ import "../js/functions.js" as Functions
 Page {
     id: imagePage
     allowedOrientations: Orientation.All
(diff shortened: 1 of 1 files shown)


qml/pages/ImagePage.qml
```diff
@@ -43,6 +44,8 @@ Page {
     property real oldCenterX;
     property real oldCenterY;
 
+    property bool imageOnly
```
monich, line 47: I could eliminate unnecessary semicolons in other places for consistency ))

Wunderfitz: No need to tease me all the time...

Wunderfitz:
Fine with me...

PR #63, opened 2020-10-12
Update Russian translation

monich:
Added the missing ones + a few tweaks, let's see how those look in the UI.


Wunderfitz:
Thx again!

PR #68, opened 2020-10-14
Fix the app icon size on the About page in landscape

monich:
It's ridiculously large (because its parent is the Column rather than Page)


Wunderfitz:
Thx, LGTM

PR #78, opened 2020-10-17
Tweaked InitializationPage UI

monich:
After having gone through initialization steps on a tablet, did a few adjustments to the initialization UI:

1. Only load app icon once (and adjust its size)
2. Removed come unused ids and attributes
3. EnterKey magic
4. Use FadeAnimation

Please take a look.


Wunderfitz:
Works here smoothy and looks great, so let's merge it! Thank you very much!

PR #79, opened 2020-10-17
Adjusted app icon size on the about page (again)

monich:
It was still too large on a 4:3 tablet screen in landscape, IMO.


Wunderfitz:
Need to check it out in detail on my tablet, but looks reasonable. Let's merge it, thanks!

PR #80, opened 2020-10-18
Notification feedback control

monich:
I find it pretty annoying to hear "bzzzz" every few seconds when a new message arrives to an active chat. Besides, running the vibra motor too often isn't good for battery life. This change introduces 3 options:

1. All events (default)
2. Only new events
3. No feedback at all

Notifications in the events view should still appear no matter what, it's only vibra and sounds that are controlled by this switch.


Wunderfitz:
Good idea, seems to work properly, let's do it! Thanks!

PR #81, opened 2020-10-18
Slightly refactored StickerPreview

monich:
I just love stickers :)

1. Removed mouse area stealing mouse events and not allowing context menu to appear when sticker is long-tapped
2. Don't load placeholder image right away, most of the time (pretty much always) they get very quickly replaced with the actual images which doesn't look great
3. Parent already provides enough paddings, there's no need to more
4. Simplified layout rules
5. Use `FadeAnimation` for animating opacity
6. `QtGraphicalEffects` import is not needed
7. Renamed `singleImage` into `stickerImage`

### DIFF
--- a/qml/components/StickerPreview.qml
+++ b/qml/components/StickerPreview.qml
@@ -17,37 +17,29 @@
     along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
 */
 import QtQuick 2.5
-import QtGraphicalEffects 1.0
 import Sailfish.Silica 1.0
 
 Item {
-
-    id: stickerPreviewItem
-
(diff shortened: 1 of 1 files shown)


Wunderfitz (changes requested):
I also love stickers :) - however, something seems to be wrong with the sizing, see the screenshot of my device...

image

monich:
Yes, indeed - if I enable "Show stickers as images" I get that too. My bad, will take another look.

monich:
I think I fixed that. Please try again!

Wunderfitz (approved):
Looks good now, thanks for the change! :)

PR #87, opened 2020-10-19
Update InitializationPage.qml

GNUuser:


qml/pages/InitializationPage.qml
```diff
@@ -147,7 +147,7 @@ Page {
 
                 TextField {
                     id: phoneNumberTextField
-                    placeholderText: "Use the international format, e.g. +4912342424242"
+                    placeholderText: qsTr("Use the international format, e.g.") + "+4912342424242"
```
jgibbon, line 150: If you want to split the string, maybe ".arg()" is the better way?
```suggestion
                    placeholderText: qsTr("Use the international format, e.g. %1").arg("+4912342424242")
```

GNUuser: ok

Wunderfitz:
New variant is OK, thanks for discovering and fixing that!

PR #92, opened 2020-10-19
implement chat info page

### LINKED ISSUE #27: feature request: Look at user profile
I like to look at the user profiles (sometimes). Clicking on the user icon should show the user profile. Very nice would be the possibility to scroll thru all user avatar pictures.

jgibbon:
This PR implements a Chat Info page. Fixes #27.
It does change/add rather a lot – as I already mentioned: sorry for that. 
Missing stuff just kept piling on, but I tried to limit myself at some point so there's still quite a lot missing.
One Example: The tab view was originally planned for displaying media content alongside the members/groups list, but that would've taken even more effort, including searches in c++ which I wasn't quite comfortable implementing by myself. So I used it for group settings for now. 
I think it could make sense to move those settings into a dedicated page when/if media tabs are added at some point.

Also, I'm almost sure some bugs crept in, please be sure to tell me if something surfaces.

## Added Features
 - PullDownMenu: Has "leave chat" and "mute/unmute" MenuItems
 - Profile picture area: If scrolled up (and a picture is available), the info page shows the big version of the profile picture. It can be clicked to go to the ImagePage. If the chat type is PrivateChat and there are multiple images, they are shown in a "Carousell".
 - Chat title: If the user is a group member with the corresponding permissions, the chat title is also (otherwise it's only in the PageHeader) shown in the content area, together with a button to change the text.
 - Chat info text/user bio: If it's the user's own page or is a group member with the corresponding permissions, a button to change the text is shown.
 - If available, the phone number is shown as a link.
 - If available, the invite link is shown as a link with a button to copy it.
 - Tab view: On the bottom of the page, a tabbed view for further stuff is added. For groups, the members are displayed in a list. For users, a list of common groups is shown. If the current user has certain permissions, a second tab item is shown to set group chat permissions, chat history visibility and slow mode (is that even supported in the ChatPage?). I also "left in" (commented out) a debug pane to quickly take a look at some values.

## Changed Things
 - The ListItems from the OverviewPage are now in a separate qml file to enable re-use.
 - ProfileThumbnail has minor changes to enable non-changing radius for different dimensions.
 - functions.js updateChatPartnerStatusText was moved here. Also, support for messageChatChangeTitle was added.
 - ChatListModel: To easily filter the lists from qml, a get(int row) method (like in listModel) was added.
 - ChatModel: messagesReceived signal was changed to include the total count. beginResetModel/endResetModel was added to initialize() to propagate changes to qml (visible with empty/new chats)
 - TDLibReceiver: messagesReceived signal was changed to include the total count. Also, a whole bunch of td-methods/signals were added.
 - TDLibWrapper: messagesReceived signal was changed to include the total count. Obviously loads of methods/signals were added here, as well.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -34,13 +34,27 @@ SOURCES += src/harbour-fernschreiber.cpp \
 
 DISTFILES += qml/harbour-fernschreiber.qml \
     qml/components/AudioPreview.qml \
+    qml/components/ChatListViewItem.qml \
     qml/components/DocumentPreview.qml \
     qml/components/ImagePreview.qml \
     qml/components/InReplyToRow.qml \
     qml/components/LocationPreview.qml \
     qml/components/StickerPicker.qml \
+    qml/components/PhotoTextsListItem.qml \
     qml/components/WebPagePreview.qml \
+    qml/components/chatInformationPage/ChatInformationEditArea.qml \
+    qml/components/chatInformationPage/ChatInformationProfilePicture.qml \
+    qml/components/chatInformationPage/ChatInformationProfilePictureList.qml \
+    qml/components/chatInformationPage/ChatInformationTabItemBase.qml \
+    qml/components/chatInformationPage/ChatInformationTabItemDebug.qml \
+    qml/components/chatInformationPage/ChatInformationTabItemMembersGroups.qml \
+    qml/components/chatInformationPage/ChatInformationTabItemSettings.qml \
+    qml/components/chatInformationPage/ChatInformationTabView.qml \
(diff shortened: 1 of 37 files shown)


jgibbon:
Small addition: It doesn't implement "Clicking on the user icon should show the user profile" yet #27

Wunderfitz (approved):
Wow, this is really a lot. Looks clean and easily understandable. Seems to work very well and also introduces some nice eye candy. Thank you very much!

Found some issues, but let's address these things iteratively. Makes it also easier as the last release has just been there, so there is no huge pressure there. 

--> Approved, will merge.

Issue #70, opened 2020-10-14
[feature request] option to mark chats as read or unread


lal883:
Would be nice to have a feature to mark all the new messages in an unread conversation as "read".

Wunderfitz:
Mark as "unread" is not supported by Telegram (as far as I know). Mark as read was implemented in v0.4 by @jgibbon.

yusssufff:
Hi!
Well both in the android app or on the Sailfish's "Depecher" app, each chat can be ticked as either "read" or "unread"...

PR #94, opened 2020-10-19
implement "mark all messages as read" MenuItem

jgibbon:
Contributes to #70 
Adds a simple forced "ViewMessage" entry for chats with unread messages.

Because the Title of #70 implies it: This doesn't mark anything as unread, though, and does not use toggleChatIsMarkedAsUnread at all.


Wunderfitz:
Still need to check whether an option "mark message as unread" would really work - at least I haven't found a dedicated option for that. But let's see - if that really doesn't work we can close #70 then. Thank you very much, LGTM!

PR #101, opened 2020-10-23
implement polls

jgibbon:
Originally, I wanted to start with displaying polls only. 
But after that was done, it turned out that creating them wouldn't take long, as well. 
The message component isn't as pretty as it could be, but I focussed on making the most of the available space.
I tried lots of variants for Buttons, IconButtons, clickable Text and think I've come to a viable compromise.
There's also a bit of a workaround to add entries to the message context menu, which ideally should get optimized at some point (uses "internal" properties).
Also, because the message items get replaced completely, I didn't make any transitions between states/results/…

Another thing that isn't really working is showing (or setting) hints for the right answer in quiz mode. 
The "explanation" member of pollTypeQuiz seems to be missing in tdlib.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -39,6 +39,7 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/components/ImagePreview.qml \
     qml/components/InReplyToRow.qml \
     qml/components/LocationPreview.qml \
+    qml/components/PollPreview.qml \
     qml/components/StickerPicker.qml \
     qml/components/PhotoTextsListItem.qml \
     qml/components/WebPagePreview.qml \
@@ -60,6 +61,8 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/pages/InitializationPage.qml \
     qml/pages/OverviewPage.qml \
     qml/pages/AboutPage.qml \
+    qml/pages/PollCreationPage.qml \
+    qml/pages/PollResultsPage.qml \
     qml/pages/SettingsPage.qml \
     qml/pages/VideoPage.qml \
     rpm/harbour-fernschreiber.changes.in \
--- a/qml/components/PollPreview.qml
+++ b/qml/components/PollPreview.qml
@@ -1,0 +1,293 @@
+/*
+    Copyright (C) 2020 Sebastian J. Wolf and other contributors
+
+    This file is part of Fernschreiber.
+
+    Fernschreiber is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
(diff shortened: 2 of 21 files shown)


monich:
Hehe, this code is using 3 different ways to animate opacity:

```qml
Behavior on opacity { FadeAnimator {} }
Behavior on opacity { NumberAnimation {}}
Behavior on opacity { PropertyAnimation {duration: 500; easing.type: Easing.InOutCubic}}
```
Need more ways of doing the same thing? I can suggest one more!
```qml
Behavior on opacity { FadeAnimation {} }
```
And I think it's actually the best one. Even though Animators are supposed to be better from the performance prospective, they sometimes get stuck and keep on running in the background, burning CPU cycles even when the screen is off, quickly draining the battery. I've never experienced that with Animations.

jgibbon:
Thanks, good catch and much appreciated. 
I will push some changes in the next days.
Cheers!

monich:
Another general remark: I personally prefer (and AFAIU that's the way blessed by Qt) to build and maintain native models and use QML only for UI. That kind of approach provides more compile time checking and performs better at run time.

However, it's a fairly different way of thinking - and if something stops you from taking that route then so be it. And yes, I know some people like writing pure QML for the reasons (possibly aesthetic) which I haven't been able to comprehend.

jgibbon:
While I could ramble at length about a different view of "front end" vs "back end" and how this just creates a simple and at all times small object, in all honesty it mainly boils down to how bad my c++ skills are. 
While I'm in no way against learning, reality (and mostly time constraints) push me into just getting working things done the way I can "blindly" type them down. As an example, I've needed almost a day to figure out what to change in the chatmodel to get it to reset and propagate that change to qml correctly (turns out, it was just startResetModel/endResetModel).
It's another reason I'm extremely glad to have someone with a deeper understanding look over my results – thank you for taking the time. I've already learned a few things contributing here.

I've now changed the Animations according to your recommendation. Because FadeAnimation is basically just a preset for "NumberAnimation", I've changed some others to NumberAnimation, as well. (Don't know why it's better than PropertyAnimation, actually, but… I don't think it can hurt.)
Also I've changed the c++ part after reading #103. 

Thanks again!

monich:
FYI, begin/endResetModel should be avoided because that completely resets the view state (most importantly, currentItem and the scrolling position). Good UI should use begin/endInsert/Move/RemoveRows blocks and dataChanged signals to propagate changes to the UI. That allows the view to behave a lot nicer and if you get creative, even apply transitions to animate the changes in all kinds of cool ways.

Wunderfitz:
Wow, thank you so much. I'm still not at 100% speed (dev machine migration) and also have other duties today, but I'll review and merge in due time. :)

Wunderfitz:
OK, initial tests work fine, code looks understandable, I'll merge it! Thank you very much! fixes can still come later...

PR #103, opened 2020-10-23
TDLibWrapper cleanup

monich:
Pass primitive data as a const reference just doesn't make any sense.

Reference (essentially a pointer) can't be smaller than a primitive type, except for 64-bit numbers on a 32-bit platform. In terms of performance, an additional indirection can't be better than picking the value directly from the stack, let alone a register.

There's no reason to do that, really. I wonder why it was done that way in the first place.


monich:
Signals can be connected directly to signals - that's more efficient than passing them though a dummy slot.

Wunderfitz:
Well, I introduced that to make it easier in the future to change the data if necessary, but in the end, you're absolutely right. Such intermediate steps can also be introduced again later if we really need it. I'll review and merge when I have more time...

Wunderfitz:
Thanks for the cleanup! Looks good, let's merge it...

PR #104, opened 2020-10-24
Update harbour-fernschreiber-fi.ts

jorm1s:
New strings added, typos fixed & small changes to better comply with the Sailfish OS Finnish style guide.

There also was a problem with the cover page phrase 'in * chats' (which is split into two different strings), as the 'in' couldn't be removed simply by leaving it blank. Worked around this by adding a zero-width-space along with a translator comment explaining this. This works, but feels a bit hacky. A much nicer way to achieve this would be to have the phrase as one string and the number embedded within it (as %1), but I'm not sure if this would allow the number to be of different font size than the words as they are currently...


Wunderfitz:
Thanks for the hint and the corrections - maybe it can be solved differently... But for now we probably need to stick to the 'hacky' option.

PR #105, opened 2020-10-25
Tweaked ImagePage behavior

monich:
Introduced a second flickable - one is strictly vertical, responsible for showing the pulley menu (which becomes non-interactive in the image-only mode), and the other one for moving the viewport around zoomed image. Making MouseArea a child of PinchArea seems to allow it to receive mouse events more reliably, adding an empty onDoubleClicked handler introduces a delay before the actual click and onClicked (so that it doesn't get invoked too often if you're tapping like crazy). Also added a scroll decorator, even though it's pretty hard to see.

It's still not easy to pull down the menu or flick the page back without scrolling the viewport to the upper left corner of the zoomed image, but I think this is way better than before.

### DIFF
--- a/qml/pages/ImagePage.qml
+++ b/qml/pages/ImagePage.qml
@@ -91,11 +91,9 @@ Page {
     }
 
     SilicaFlickable {
-        id: imageFlickable
         anchors.fill: parent
-        clip: true
-        contentWidth: imagePinchArea.width;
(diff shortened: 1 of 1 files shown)


Wunderfitz:
Yep, seems to work smoothly, thanks for that. And I really like the scroll decorator! :)

PR #106, opened 2020-10-26
use svg icons

### LINKED ISSUE #102: Use svg icons
Fixed-sized png icons should be replaced with their svg equivalents to keep the UI resolution independent (those pngs are most likely generated from svgs anyway). Even if pngs get resized to a reasonable size on the screen, they won't look as great as properly rendered svgs when screen resolution doesn't match the expected one.

jgibbon:
Closes #102.

Doesn't use Shader/HighlightImage for background/placeholder, since that neither matched the previous look nor meet my visual expectations at all. I created both previously available color variations as svg accordingly.
Implementation of the background images has been outsourced to a dedicated qml file to reduce duplication and ensure the same layout. Padding is now "medium" everywhere (stickers had none, audio had "small").

For the Icons, some (Highlight-)Image+MouseArea combinations have been changed to IconButton.
Regarding videos, I'm not actually sure if the missing "highlight" on the fullscreen icon was intentional. An issue with videos possibly having the same color as the icon remains basically the same. Nonetheless, if this isn't desired, I'll revert to just using the white image.

The sticker icon actually does look a bit differently now – it's a bit smaller. I've redrawn it from scratch but mostly kept the dimensions of icon-m-other (which doesn't look good when zoomed in, btw). Perhaps I actually did made a mistake drawing the previous one. Either way, I think it fits well enough.

The notification icon remains png, but I managed to reduce its file size by ~60% without visibly impacting it much imho.

This may be the longest commit message I've ever written.


Wunderfitz:
Wow, really cool. Thank you very much! :)

PR #108, opened 2020-10-27
Fix emoji paths for chat information Items

jgibbon:
It didn't really work in most places and the others did only work due to a bad workaround.


Wunderfitz:
Good catch, thanks! :)

PR #109, opened 2020-10-28
Notifications

monich:
I noticed bzzzz when a group was being marked as read by another client (even though I had "Only new events" selected in the settings), and in the process of fixing that, made a bunch of other changes.

Summary:

1. Don't post unwanted feedback for removed notifications
2. Don't turn off PatternCommunicationIM pattern when some (but not all) notifications are removed
3. Don't keep empty notification groups in notificationGroups map
4. Pre-allocate commonly used QStrings
5. Don't pass primitive types by const references to NotificationManager
6. Use const objects when appropriate to make sure that they don't get unnecessarily detached (i.e. duplicated)

I tried to split unrelated changes into different commits.

### DIFF
--- a/src/notificationmanager.cpp
+++ b/src/notificationmanager.cpp
@@ -136,27 +136,27 @@ void NotificationManager::handleChatDiscovered(const QString &chatId, const QVar
     this->chatMap.insert(chatId, chatInformation);
 }
 
-void NotificationManager::handleNgfConnectionStatus(const bool &connected)
+void NotificationManager::handleNgfConnectionStatus(bool connected)
 {
     LOG("NGF Daemon connection status changed" << connected);
 }
 
-void NotificationManager::handleNgfEventFailed(const quint32 &eventId)
+void NotificationManager::handleNgfEventFailed(quint32 eventId)
 {
     LOG("NGF event failed, id:" << eventId);
 }
 
-void NotificationManager::handleNgfEventCompleted(const quint32 &eventId)
+void NotificationManager::handleNgfEventCompleted(quint32 eventId)
 {
     LOG("NGF event completed, id:" << eventId);
 }
 
-void NotificationManager::handleNgfEventPlaying(const quint32 &eventId)
+void NotificationManager::handleNgfEventPlaying(quint32 eventId)
 {
     LOG("NGF event playing, id:" << eventId);
 }
 
-void NotificationManager::handleNgfEventPaused(const quint32 &eventId)
+void NotificationManager::handleNgfEventPaused(quint32 eventId)
 {
     LOG("NGF event paused, id:" << eventId);
 }
@@ -232,7 +232,7 @@ QString NotificationManager::getNotificationText(const QVariantMap &notification
     return FernschreiberUtils::getMessageShortText(notificationContent, false);
 }
 
-void NotificationManager::controlLedNotification(const bool &enabled)
+void NotificationManager::controlLedNotification(bool enabled)
 {
     static const QString PATTERN("PatternCommunicationIM");
     static const QString ACTIVATE("req_led_pattern_activate");
--- a/src/notificationmanager.cpp
+++ b/src/notificationmanager.cpp
@@ -19,7 +19,6 @@
 
 #include "notificationmanager.h"
 #include "fernschreiberutils.h"
-#include <nemonotifications-qt5/notification.h>
 #include <sailfishapp.h>
 #include <QDebug>
 #include <QListIterator>
@@ -34,9 +33,7 @@ namespace {
     const QString TYPE("type");
     const QString ID("id");
     const QString CHAT_ID("chat_id");
-    const QString IS_SILENT("is_silent");
     const QString IS_CHANNEL("is_channel");
-    const QString REPLACES_ID("replaces_id");
     const QString TOTAL_COUNT("total_count");
     const QString DATE("date");
     const QString TITLE("title");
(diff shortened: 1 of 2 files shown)


src/notificationmanager.cpp
```diff
@@ -67,25 +99,31 @@ void NotificationManager::handleUpdateActiveNotifications(const QVariantList not
...
-    notificationGroup.insert("is_silent", notificationGroupUpdate.value("is_silent"));
-    notificationGroup.insert("total_count", notificationGroupUpdate.value("total_count"));
+    const QVariantList addedNotifications = notificationGroupUpdate.value(ADDED_NOTIFICATIONS).toList();
+    const QVariantList removedNotificationIds = notificationGroupUpdate.value(REMOVED_NOTIFICATION_IDS).toList();
 
-    QVariantMap activeNotifications = notificationGroup.value("notifications").toMap();
+    // Determine if we need a bzzzzz
+    const AppSettings::NotificationFeedback feedbackStyle = appSettings->notificationFeedback();
+    const bool needFeedback = (feedbackStyle != AppSettings::NotificationFeedbackNone &&
+        (!addedNotifications.isEmpty() || feedbackStyle == AppSettings::NotificationFeedbackAll));
```
Wunderfitz, line 120: That seems to cause an NGF action (bzzzz) each time a notification group is updated. On my device every time, I read a message in a chat that is not muted and if there are still unread messages in unmuted chats.

src/notificationmanager.cpp
```diff
@@ -95,39 +133,51 @@ void NotificationManager::handleUpdateNotificationGroup(const QVariantMap notifi
...
-        QVariantMap activeNotification = this->sendNotification(chatId, addedNotification, activeNotifications);
-        activeNotifications.insert(activeNotification.value("id").toString(), activeNotification);
+        const QVariantMap addedNotification = addedNotificationIterator.next().toMap();
+        const QVariantMap activeNotification = this->sendNotification(chatId, addedNotification, activeNotifications, needFeedback);
+        activeNotifications.insert(activeNotification.value(ID).toString(), activeNotification);
+    }
+
+    if (needFeedback) {
+        ngfClient->play(NGF_EVENT);
     }
```
Wunderfitz, line 159: Why did you separate the NGF play event from the actual action to send a new notification? Seems to be inconsistent and possibly triggers too many bzzzzs...

Wunderfitz (changes requested):
Thanks for the changes, really appreciated! 

However, according to my tests there are some regressions - too many NGF events in certain cases, probably depending on the configuration. Moreover, I don't really see the need to separate the sending of notifications and the triggering of the NGF event. That might cause inconsistencies...

src/notificationmanager.cpp
monich: My intention was actually to reduce number of bzzz - it used to be inside `sendNotification` (which can be called more than once), moving bzzz outside can't trigger more, only less.

src/notificationmanager.cpp
monich: I'll take another look at it.

monich:
OK, you forced me to dig deeper into this notifications stuff. Here we go ))

Wunderfitz:
:D Thanks! Will review now...

monich:
In addition to what was done in the first version of this pull request, this one also attaches to the existing notifications at startup (to avoid duplicates) and tracks their state by keeping `Notification` object alive (it's receiving D-Bus notifications from the events view and updates its `replacesId` attribute accordinly).

monich:
Chat title updates ate handled too and get propagated to the events view.

src/notificationmanager.cpp
monich: Actually, I ended up moving it back to the method which actually posts the notifications...

src/notificationmanager.cpp
```diff
@@ -53,176 +142,252 @@ NotificationManager::NotificationManager(TDLibWrapper *tdLibWrapper, AppSettings
...
 
+    nemoNotification->setBody(notificationBody);
+    nemoNotification->setSummary(chatInformation ? chatInformation->title : QString());
     if (needFeedback) {
-        nemoNotification.setCategory("x-nemo.messaging.im");
-        ngfClient->play("chat");
+        nemoNotification->setCategory(NOTIFICATION_CATEGORY);
+        nemoNotification->setPreviewBody(nemoNotification->body());
+        nemoNotification->setPreviewSummary(nemoNotification->summary());
```
Wunderfitz, line 382: I probably remove them for now as I explicitly didn't add previewBody/Summary in the original version. They cause notification popups even when I'm in the respective chat. I might add a small 

if (not in chat that receives notification) then (set previewBody/Summary to body() and summary()) else (set previewBody/Summary to QString())

in the future.

Wunderfitz (approved):
Works great, thank you very much! I might add a small thing as mentioned in the other comment, but nothing that blocks the merge....

src/notificationmanager.cpp
monich: If you don't set those explicitly, they won't be automatically updated together with body/summary and you get a mismatch.

Wunderfitz: Yep, therefore 86ac99004d170dee622fefc7f7bcebb5bb11d562 sets them now explicitly to an empty string...

monich: But yes, if you're already in this chat, a number in a circle in the bottom right corner would be a far more appropriate notification mechanism then a popup.

PR #111, opened 2020-10-28
Fix two issues in PollPreview

jgibbon:
- The new emoji basePath workaround also works for TextSwitch content.
 - Checking which context menu entries get added for the poll preview didn't quite work nicely "OnCompleted".


Wunderfitz:
Thx!

PR #114, opened 2020-10-30
Reduce ChatPage.qml jit compile time

jgibbon:
First of all: Take all measurements I mention with a grain of salt – all of them are rough and not necessarily measured more than a few times. All times were measured on an Xperia X run via SDK.

Visiting a chat page can take a long time, especially before the qml is cached by the engine.
When opening it for the first time after application launch, it sometimes takes >1000ms from onClicked (OverviewPage) to Component.OnCompleted (Chatpage).
Subsequent activations take roughly 470-480ms.

With these changes, I was able to reduce these times to ~450ms for the first, ~100ms for subsequent activations of the ChatPage on my test device.

There are some run-time optimizations possible (some of which I possibly created with this), I did not focus on those at all.

Things changed:
- The components for displaying extra content to a message are (mostly) gone and replaced by a single Loader. This Loader does not use sourceComponent to trade the initial compilation boost for a neglegible bit of runtime penalty.
- Connections were consolidated
- I was surprised how costly the inclusion of the RemorseItem was (compiling ~75ms, initializing up to ~20ms for every delegate). So I traded a bit for a compromise. deleteMessageRemorseItem is now defined on the appWindow level, where it gets a bit mitigated by the animations at application start. Also, only one deletion at a time is now possible. We can easily revert this change, but I thought it worthwhile despite its drawbacks.
- profileThumbnailComponent is now defined directly as sourceComponent, removing the need for its id. Probably didn't do anything.
- InReplyToRow had width: parent.width, so I removed horizontalCenter. Also probably didn't change compilation time at all.
- Another compromise I was willing to take – your opinion may differ: The PickerPages took ages (~200ms) to just parse/compile inside those Components, so I replaced them with the "string notation" of pageStack.push. Drawback: The first time a picker gets activated, you'll see how slow it is. Subsequent activations aren't that bad – also for the other pickers.

### DIFF
--- a/qml/components/AudioPreview.qml
+++ b/qml/components/AudioPreview.qml
@@ -24,15 +24,18 @@ import "../js/functions.js" as Functions
 Item {
     id: audioMessageComponent
 
-    property variant audioData;
+    property ListItem messageListItem
+    property variant rawMessage: messageListItem.myMessage
+
+    property variant audioData: ( rawMessage.content['@type'] === "messageVoiceNote" ) ?  rawMessage.content.voice_note : ( ( rawMessage.content['@type'] === "messageAudio" ) ? rawMessage.content.audio : "");
     property string audioUrl;
     property int previewFileId;
     property int audioFileId;
-    property bool onScreen;
+    property bool onScreen: messageListItem.page.status === PageStatus.Active
     property string audioType : "voiceNote";
 
     width: parent.width
-    height: parent.height
+    height: width / 2
 
     function getTimeString(rawSeconds) {
         var minutes = Math.floor( rawSeconds / 60 );
--- a/qml/components/DocumentPreview.qml
+++ b/qml/components/DocumentPreview.qml
@@ -26,7 +26,10 @@ Item {
     width: parent.width
     height: Theme.itemSizeLarge
 
-    property variant documentData;
+    property ListItem messageListItem
+    property variant rawMessage: messageListItem.myMessage
+
+    property variant documentData: rawMessage.content.document
     property bool openRequested: false;
 
     Component.onCompleted: {
--- a/qml/components/ImagePreview.qml
+++ b/qml/components/ImagePreview.qml
@@ -24,8 +24,13 @@ Item {
 
     id: imagePreviewItem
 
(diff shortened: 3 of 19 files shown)


Wunderfitz:
Thanks a lot, just tried it on my test device and the improvement is really impressive. The only downside is that the precalculated height of the lazy-loaded components is gone - that leads to jumping in the list during scrolling. I'd like to avoid that... Are the dedicated loaders really that heavy? Maybe the lazy loading of the remorse item and of the pickers also already do a good job. I'll probably experiment tomorrow myself a bit.

monich:
Consider using [`Remorse.itemAction`](https://sailfishos.org/develop/docs/silica/qml-sailfishsilica-sailfish-silica-remorse.html/#itemAction-method) and [`Remorse.popupAction`](https://sailfishos.org/develop/docs/silica/qml-sailfishsilica-sailfish-silica-remorse.html/#popupAction-method) - those create remorse components on demand. And in general, use Components (rather than live instances) for heavy things whenever API permits it - that postpones instantiation of the actual component until it's actually needed.

jgibbon:
@Wunderfitz as discussed privately, I may just add the different height presets to the one Loader. The loaders weren't the problem, as such, btw. It was the components.

@monich Nice, I'll look into it.
Yeah, in this case even the components added up to a lot of qml compiling time. 

I can't say if I get the chance to work on this tonight but I'll certainly revisit both aspects mentioned above. :)

jgibbon:
I have re-introduced the height preset(s) and used Remorse.itemAction, thanks again for those suggestions!

Funnily enough I've got a feeling that we now could get away with either setting the delegateComponentLoadingTimer.interval really tiny or even removing that Timer alltogether.

We'll still need a onHeightChanged-solution for extra content with variable/changing height at some point, but that's a topic we already discussed in another issue and largely unconnected to this one.

cheers

Wunderfitz:
Thank you very much. Seems to work with the latest changes and the small adjustments for the pickers... ;) Let's merge it!

jgibbon:
Oh, must have missed that, sorry.

PR #115, opened 2020-10-31
Update harbour-fernschreiber-it.ts

iamnomeutente:
I don't know why I have a problem with the ellipsis and why also the commit of 27th October appears here. Sorry, I hope this is OK anyway.


Wunderfitz:
Grazie!

PR #116, opened 2020-10-31
Copy messages to clipboard

### LINKED ISSUE #15: copy & paste from chat
Please enable to extract parts of chats (like URLs) by copy n paste!

chstem:
This PR adds an item to the message context menu allowing to copy the whole message to the clipboard.
This partially resolves #15

### DIFF
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -557,6 +557,12 @@ Page {
                             }
                             MenuItem {
                                 onClicked: {
+                                    Clipboard.text = messageText.text
+                                }
+                                text: qsTr("Copy Message to Clipboard")
+                            }
+                            MenuItem {
+                                onClicked: {
                                     deleteMessageRemorseItem.execute(messageListItem, qsTr("Deleting message"), function() { tdLibWrapper.deleteMessages(chatInformation.id, [ display.id ]); } );
                                 }
                                 text: qsTr("Delete Message")
--- a/translations/harbour-fernschreiber-de.ts
+++ b/translations/harbour-fernschreiber-de.ts
@@ -278,6 +278,10 @@
         <translation>Nachricht bearbeiten</translation>
     </message>
     <message>
+        <source>Copy Message to Clipboard</source>
+        <translation>Nachricht in die Zwischenablage kopieren</translation>
+    </message>
+    <message>
         <source>edited</source>
         <translation>bearbeitet</translation>
     </message>


jgibbon:
Wow, didn't know it was so easy. I should remove my workaround elsewhere soon.

jgibbon:
There is one small issue, though:
When copying `messageText.text`, you'll not necessarily get the thing you want. It may be empty or it may contain html-markup (for example emojis) specific to Fernschreiber.

Perhaps you could make use of `Functions.getMessageText(display, true, messageListItem.isOwnMessage)` (note the `true`,  to basically return the short preview text for the overview page).
But keep in mind that some of that output would only make sense if you prepended `!messageListItem.isOwnMessage ? Functions.getUserName(messageListItem.userInformation : qsTr("You")` to the copied text.

But it's a really great start and I'd guess @Wunderfitz will merge it with pleasure after this small issue is adressed. Thank you!

cheers!

chstem:
Thanks for the feedback. I have two questions about that:

1. You are vaguely describing some special cases. Could you give more specific examples?
1. You suggested to use `simple=true`. Why would someone want to copy the short preview only? I think this should be `simple=false`.

jgibbon:
I didn't describe my thoughts clearly and my answer was a bit confusing because I rushed it a bit, sorry for that. 
Both answers to you questions lie inside the [`getMessageText`](https://github.com/Wunderfitz/harbour-fernschreiber/blob/72315312d6370b7d1e4179db13439fa827a70fda/qml/js/functions.js#L26) function. 
1. There you can find several examples like sent a picture/sent a video/sent an animation/… – many of those are only returned if the "special content" has no caption/title/file name.
2. I overly simplified my description to be not quite correct in the end: The "short preview text" as I called it does not necessarily mean it's shortened at all. Simple in this case just means "not the output for the chat page", where it actually often just is an empty string.

cheers!

chstem:
Ah, I see. Thanks for the clarification.

For a text message this would remove all the formatting text, which may or may not be desired by the user. But I think no formatting would be the better choice.

I am not sure what the expected behavior for non text messages would be. For example for images, currently the caption is copied, which is not really useful. Question: Is there a use case for copying image messages (or other types) to the clipboard? Maybe we just show the clipboard option for text messages only, and extend to other types as required.

Wunderfitz:
Thanks for the PR - should have done that earlier as it's really not that hard. However, I'd like to avoid using `messageText.text` as it's HTML markup (img tags for emojis, links, HTML tags for < > etc.). You've had the discussion already. As there is anyway a conflict that needs to be resolved, I'll correct that and merge afterwards.

PR #117, opened 2020-10-31
Move chatTypeFromString from ChatListModel to TDLibWrapper

monich:
So that it can be used elsewhere (e.g. from NotificationManager).


monich:
And I have an update for https://github.com/Wunderfitz/harbour-fernschreiber/pull/109 which depends on this one

Wunderfitz:
Sure, let's merge it!

PR #118, opened 2020-10-31
Remove clipboard workaround

jgibbon:
Turns out, I somehow forgot about the Clipboard type while implementing the info page. 
Thanks @chstem


Wunderfitz:
:)

PR #119, opened 2020-10-31
clean up qml imports

jgibbon:
This changes "old" imports from QtQuick (newest supported is 2.6) and QtMultimedia (newest 5.6).
Also removes QtMultimedia & QtGraphicalEffects from some qml files where I could not identify usage.


Wunderfitz:
Thanks for that - leftovers of too much copy and paste. ;)

PR #121, opened 2020-11-01
Don't animate ChatPage transition when notification is tapped

monich:
I find that animation unnecessary and inappropriate but to a certain degree it's a matter of taste...


Wunderfitz:
You're absolutely right. I probably simply forgot that. Moreover, thanks for simplifying the page parameter for pageStack.pop(). No idea why I missed that...

PR #126, opened 2020-11-02
Set initial scroll position for ChatPage

### LINKED ISSUE #1: Open Chat: Positioning not exact
When opening a new chat, the list is not always correctly positioned at the last read message. Seems to be related to the list view items' height that is dynamic and depends if there is a picture/video/sticker etc.. There is another option, but that is much slower and delays the opening of the chat... Let's see how it can be corrected. ;)

jgibbon:
Fixes #1 (if I haven't missed something) and uses a simplified version of the solution discussed in that issue. `ListView.ApplyRange` does work better than `ListView.StrictlyEnforceRange` for `contentHeight` < `height` and doesn't (as far as I've seen) interfere with manual scrolling, so the proposed `Timer` wasn't necessary. It also works well with chats containing "variable height" messages like polls. (Those still jump when scrolling up manually, of course.)

A workaround for quickScroll guessing wrong (scrolling down) was also added.

I did not notice any of the crashes you mentioned, but since those were quite elusive, I obviously can't be sure they won't happen.

Also, a new bool `manuallyScrolledToBottom` property was added to the `SilicaListView` to keep scrolling down when new messages arrive. 
Those messages will, as of now, only be marked as read after a manual scroll action – if Fernschreiber isn't visible, this prevents automatically "reading" and missing messages. 

Possible enhancements:
 - Mark those as read if the app is running in the foreground to spare users the manual scrolling.
 - Perhaps marking messages as read could generally be done with a `Timer` (also when scrolling) to give users a chance to see what's happening. This would be another way to adress the "they effectively just blink" part from #23. The background color could also get a transition to be a bit more "fluid".

IMHO both of those would make sense. What do you think?

### DIFF
--- a/qml/components/ProfileThumbnail.qml
+++ b/qml/components/ProfileThumbnail.qml
@@ -106,7 +106,6 @@ Item {
                 height: parent.height - Theme.paddingSmall
                 anchors.centerIn: parent
                 source: profileThumbnail.photoData.local.path
-
                 fillMode: Image.PreserveAspectCrop
                 autoTransform: true
                 asynchronous: true
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -270,18 +270,13 @@ Page {
             console.log("[ChatPage] Messages received, view has " + chatView.count + " messages, setting view to index " + modelIndex + ", own messages were read before index " + lastReadSentIndex);
             if(totalCount === 0) {
                 console.log("[ChatPage] actually, skipping that: No Messages in Chat.");
-
                 chatView.positionViewAtEnd();
                 chatPage.loading = false;
                 return;
             }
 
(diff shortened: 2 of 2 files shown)


qml/components/ProfileThumbnail.qml
```diff
@@ -106,7 +106,6 @@ Item {
                 height: parent.height - Theme.paddingSmall
                 anchors.centerIn: parent
                 source: profileThumbnail.photoData.local.path
-
```
jgibbon, line 109: ahem, I totally planned to remove this empty line… because it's very relevant to the issue, of course.

Wunderfitz: :D

Wunderfitz (approved):
Really incredible. As said in #1 - I had this variant already at the very beginning, but it was so slow because of the animations... No crashes so far let's try it and test in real life for some days.

PR #127, opened 2020-11-02
fix rpm changelog

jgibbon:
To include the changes in the rpm it
 - needs to be renamed
 - needs to have entries in descending order
Also, I've changed the second 0.2 → 0.3

If you think that's not important at all, let me tell you: 
You're probably completely right. 
(But perhaps it saves you a bit of copypasting when releasing on openrepos.)


Wunderfitz:
Good point! :)

PR #129, opened 2020-11-04
delay viewMessage calls; animate changes

jgibbon:
I've quickly added a Timer to delay viewMessage-calls
as mentioned in #126 (and discussed privately) for your consideration.


Wunderfitz:
Great, thanks! I'd probably reduce the timeout to 1 or 1.5 seconds, but let's see how it goes...

PR #132, opened 2020-11-05
Add simple message delegate

jgibbon:
I've basically moved the existing delegate into another file as-is (no optimizations or precalculated dimensions, yet) and created a simple one for the following message types:
- messageBasicGroupChatCreate
- messageChatAddMembers
- messageChatChangePhoto
- messageChatChangeTitle
- messageChatDeleteMember
- messageChatDeletePhoto
- messageChatJoinByLink
- messageChatSetTtl
- messageChatUpgradeFrom
- messageChatUpgradeTo
- messageCustomServiceAction
- messagePinMessage
- messageScreenshotTaken
- messageSupergroupChatCreate
- messageUnsupported

This means that a message type which simply isn't implemented by Fernschreiber (not the same as messageUnsupported) will still use the normal delegate.

Also I've added descriptions for a bunch of message types.


Wunderfitz:
Thanks, looks really nice and seems to work properly. Let's merge it!

PR #133, opened 2020-11-05
replace variant with var

jgibbon:
https://doc.qt.io/archives/qt-5.6/qtquick-performance.html

> In general, "property var" should be considered to be superior to "property variant" for every use-case from QtQuick 2.0 and newer (note that "property variant" is marked as obsolete), as it allows a true JavaScript reference to be stored (which can reduce the number of conversions required in certain expressions).


Wunderfitz:
Thx, good catch. Didn't know that detail...

PR #134, opened 2020-11-06
Update harbour-fernschreiber-it.ts

iamnomeutente:
New strings and some corrections (3rd-person plural > 2nd singular)
```
<source>sent a video</source>
<comment>myself</comment>
<translation type="unfinished"></translation>
```
Is this:
1. "You" sent a video (second-person singular)?
2. "They" sent a video?
3. "I" sent a video?

I used now the 1st case (2nd singular)

### DIFF
--- a/translations/harbour-fernschreiber-it.ts
+++ b/translations/harbour-fernschreiber-it.ts
@@ -413,7 +413,7 @@
     <message>
         <source>sent a picture</source>
         <comment>myself</comment>
-        <translation>hanno inviato una foto</translation>
+        <translation>hai inviato una foto</translation>
     </message>
     <message>
         <source>sent a picture</source>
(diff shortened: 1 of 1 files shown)


Wunderfitz:
Yep, that's correct that way. In German I even used the formal you ("Sie") - so the actual case depends on the way how the "You" is translated, if there is a formal "You" in your language and if you want to use it in the translation.

iamnomeutente:
Ok, thank you.
In Italian there is a formal form, but commonly it isn't used in an app. 
It would be even better without the pronoun at all. E.g. Instead of "Tu hai inviato una foto", simply "Hai inviato una foto" but maybe it's too difficult here because you have "You" and the sentences ("sent a photo", ...) separated.

PR #135, opened 2020-11-07
qml runtime optimizations

jgibbon:
Hi @Wunderfitz, as announced, here's another one with quite a few changes. 
We discussed a few of them privately already.

On normal or fast hardware, these changes will not have a huge (visible) impact, but slow hardware (PinePhone?) should benefit a bit more – also it may save a tiny bit of power. 

From my observations, most of the larger computation chunks needed for scrolling around the chat view now actually come from places we'd expect and can possibly optimize in the future: The JS functions formatting text/time/emojis.

In Addition to the `MessageListItem`s, I made some (smaller) changes to other parts:
 - `ChatListViewItem`: moved the "chat muted" icon to the title, saving up to one call to emojify per visible chat.
 - `PollPreview`: changed the custom context menus to account for the `MessageListItem` changes (see below). No more hack (using internal silica properties) necessary.
 - `ProfileThumbnail`: relying a bit more on the qml bindings, less on custom js code. It doesn't show the Text (and switch to the image) if an image is expected, instead, it just loads it. Also added sourceSize.
 - `WebPagePreview`: added sourceSize.
 - `ChatPage`: 
   - Removed some of my scroll changes (relying on currentIndex) in favour of positionViewAtIndex, which seems to fare better with the automatic currentIndex changes.
   - Removed my opacity change of the unread messages item. It seemed like a good Idea at the time but ended up going on my nerves more times than not, especially when scrolling up. I'm not completely satisfied with this.
   - added `QtObject precalculatedValues` for the MessageListItem delegates.
  
Main changes to `MessageListItem`s
 - changed `display` to `myMessage` for simplified scope lookup
 - added one `page` property to refer to the `ChatPage` (possibly unnecessary).
 - put the `ContextMenu` inside a `Loader` and added a `ObjectModel additionalContextItems` property to easily add menu items from a delegate (see PollPreview above).
 - added properties for textColor and textAlign to reduce duplicated conditions
 - removed a call to `locationPreviewLoader` I forgot in there.
 - put `messageInReplyToRow` into a `Loader`
 - used `precalculatedValues` for most custom `width` declarations
 - replaced some anchor declarations for `messageBackground` with a `precalculatedValues` width
 - added `isUnread` property to messageBackground to reduce condition checks and assignments
 - unified two `Component.onCompleted` functions
 - removed `heightPreset` from `extraContentLoader` and put the value inside the `height` condition directly (so it saves a few cycles if not needed)

cheers!

### DIFF
--- a/qml/components/ChatListViewItem.qml
+++ b/qml/components/ChatListViewItem.qml
@@ -13,7 +13,7 @@ PhotoTextsListItem {
     property url emojiBase: "../js/emoji/"
 
     // chat title
-    primaryText.text: title ? Emoji.emojify(title, Theme.fontSizeMedium, emojiBase) + ( display.notification_settings.mute_for > 0 ? Emoji.emojify(" 🔇", Theme.fontSizeMedium) : "" ) : qsTr("Unknown")
+    primaryText.text: title ? Emoji.emojify(title + ( display.notification_settings.mute_for > 0 ? " 🔇" : "" ), Theme.fontSizeMedium, emojiBase) : qsTr("Unknown")
     // last user
     prologSecondaryText.text: is_channel ? "" : ( last_message_sender_id ? ( last_message_sender_id !== ownUserId ? Emoji.emojify(Functions.getUserName(tdLibWrapper.getUserInformation(last_message_sender_id)), primaryText.font.pixelSize) : qsTr("You") ) : qsTr("Unknown") )
     // last message
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -25,19 +25,19 @@ import QtQml.Models 2.3
 ListItem {
     id: messageListItem
     contentHeight: messageBackground.height + Theme.paddingMedium
-    property var myMessage: display
-    property var userInformation: tdLibWrapper.getUserInformation(myMessage.sender_user_id)
+    readonly property var myMessage: display
+    readonly property var userInformation: tdLibWrapper.getUserInformation(myMessage.sender_user_id)
     property QtObject precalculatedValues: ListView.view.precalculatedValues
-    property color textColor: isOwnMessage ? Theme.highlightColor : Theme.primaryColor
-    property int textAlign: isOwnMessage ? Text.AlignRight : Text.AlignLeft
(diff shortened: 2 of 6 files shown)


monich:
Midair collision)) My https://github.com/Wunderfitz/harbour-fernschreiber/pull/138 optimizes `ProfileThumbnail` too, albeit differently - by pushing file loading logic to a native object, killing the timer and all JavaScript stuff related to fetching the file. I'm not 100% sure that I tested all corner cases though, such as actually fetching the file from the network, let along failure to load the file. But the overall idea should be pretty obvious.

monich:
Another thing related to optimization - I've been repeatedly told by QML gurus that marking properties as readonly is good for performance and I have no reason to not trust those guys)) Although I can't tell why exactly it's good and how significant the effect is.

jgibbon:
> Midair collision)) My #138 optimizes `ProfileThumbnail` too, albeit differently - by pushing file loading logic to a native object, killing the timer and all JavaScript stuff related to fetching the file. I'm not 100% sure that I tested all corner cases though, such as actually fetching the file from the network, let along failure to load the file. But the overall idea should be pretty obvious.

From first glance I'd say your changes make more sense in the long run and basically include what I've done anyway. 

@Wunderfitz if #138 works to your satisfaction, feel free to merge that first. I'll happily rebase/force push then.

> Another thing related to optimization - I've been repeatedly told by QML gurus that marking properties as readonly is good for performance and I have no reason to not trust those guys)) Although I can't tell why exactly it's good and how significant the effect is.

Good point. I did try changing some to readonly in a previous attempt – I guess it saves on binding time. Totally forgot with this pr though. I'll investigate where readonly might make sense for this PR.

Wunderfitz:
Thanks for all the work, that looks really good! :) As you could see, I merged #138 as it seemed to work smoothly and IMHO it's the best option in the long run. As we are still on a 0.x, I suppose we should go for it and fix the bugs later (if there are any). So thanks @jgibbon for offering to integrate #138 in this PR. 

Happy to merge this one here as well once you say it can be reviewed!

Wunderfitz:
At least on my XA2 Ultra, the differences are obvious - awesome! Didn't find any issues, so let's merge it. Thanks again!

PR #138, opened 2020-11-08
Add TDLibFile and optimize ProfileThumbnail

monich:
Profile images seem to be loading significantly faster after moving file fetching logic to the native code and removing the artificial delay.

TDLibFile is a generic object which can hopefully be used elsewhere as an efficient replacement for JavaScript.


Wunderfitz:
Wow, thanks! And yes, the new TDLibFile can certainly be re-used also in other areas...

Seems to work properly - also on older devices such as the Jolla Phone, so let's merge it.

PR #140, opened 2020-11-11
Support for animated stickers

monich:
Support for TGS files is implemented as a [QImageIOHandler](https://doc.qt.io/qt-5/qimageiohandler.html). Rendering is done by [rlottie](https://github.com/Samsung/rlottie) and it's pulled in as a submodule. The first frame is cached, all others are constantly re-rendered, saving memory but chewing up CPU time, eating the battery and warming the device.

Sticker animation can be turned off via settings - the very same plugin works for static images too, although in this case it unnecessarily allocates one extra frame (returns the first one and then pre-fetches the second which never gets requested), but IMO that's an adequate price to pay for the simplicity of the implementation.


monich:
This also fixes sticker alignment which got broken (became always centered) at some point.

jgibbon:
Again: Impressive! And fast, as well!
As you've mentioned, it has a noticeable performance impact (my Xperia X struggles a tiny bit – obviously less so without SDK/Debug/SSH), but from a quick glance, it seems to work pretty well, you even took care to render the non-animated version nicely.
Thanks for investing your time for this – I don't think @Wunderfitz or me would have gotten this good of a result, if any.

This may be a first for a SFOS Telegram client ;)

cheers

Wunderfitz:
Thank you very much, works smoothly on the devices which I could test in my short break this noon... There's a small glitch with the sizing on the Jolla 1 (which performs really well for these animations!), but I'll correct that later. Merging this one for now. :+1:

PR #146, opened 2020-11-12
Updated Russian translation

monich:
Still incomplete, got tired ))


Wunderfitz:
Thanks! And take your time for a break - we need to awake and fit! ;)

PR #159, opened 2020-11-14
multiple message selection

jgibbon:
Selection Mode can be activated by selecting the first messages via its ContextMenu. Initially, we discussed the PullDownMenu, but I did not go that route because the actions available are dependent of which messages are selected and it just looked a bit off.
In Selection Mode, tapping a message toggles its selection state.
Selection Mode is exited if no message is selected. This happens after most actions, by long pressing any message, by pressing the X icon or by pressing the page header.

Three actions are implemented:
 - copy messages (adds sender name in front of the messages compared to the existing one)
 - delete
 - forward (this opens a chat selection dialog)

Also, I've replaced a few photoData "" fallbacks with {}

cheers

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -64,6 +64,7 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/js/functions.js \
     qml/pages/ChatInformationPage.qml \
     qml/pages/ChatPage.qml \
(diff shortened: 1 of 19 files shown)


Wunderfitz:
Awesome work, thanks! Maybe a small notification would be nice if you forward to a chat that you have no access to, but that can easily be added later. Merging...

PR #163, opened 2020-11-15
add en translation

jgibbon:
Added en translation because the default "POSIX fallback" never handles plural strings correctly.


Wunderfitz:
Thanks! :+1:

PR #167, opened 2020-11-16
Removed mutex from ChatModel

monich:
It appears to be useless, at least I couldn't figure out what's being synchronized with what

### DIFF
--- a/src/chatmodel.cpp
+++ b/src/chatmodel.cpp
@@ -22,6 +22,7 @@
 #include <QListIterator>
 #include <QByteArray>
 #include <QBitArray>
+#include <QDebug>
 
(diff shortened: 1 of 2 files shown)


Wunderfitz:
Well, I simply wanted to be cautious at the beginning in order to avoid race conditions and similar things. But of it works without, I'm fine with it. Will test it shortly and merge if I don't encounter any obvious issues...

Wunderfitz:
No suspicious things notices, let's merge. Can re-introduce the mutex in case we need it...

monich:
Introducing new entities without necessity is almost never a good idea ))

Wunderfitz:
That's so true. However, I was a little desperate at that time as I had quite a few strange issues in both the chat list and the chat model...

PR #168, opened 2020-11-16
ChatInfoPage as attached page

### LINKED ISSUE #150: Access Chat information page with a swipe
Instead to access the chat information page only by clicking on the header it would be nice also with a swipe from right to left

### LINKED ISSUE #166: Group members list is repeated several times for some groups
The group member list in the chat info page (opened by taping the top bar in any group chat) seems to be repeated several times for some groups, eg. the Fernschreiber Development group. I checked several groups and most of them had this bug, but didn't find any clear indicators which groups would or would not trigger this.

jgibbon:
fixes #150: Now basically everything is inside a loader; ChatInformationPage is added to ChatPage with pageStack.pushAttached

fixes #166: Replaces the clunky VisualItemModel in tab view and doesn't initialize multiple times.

### DIFF
--- a/qml/components/chatInformationPage/ChatInformationPageContent.qml
+++ b/qml/components/chatInformationPage/ChatInformationPageContent.qml
@@ -114,7 +114,7 @@ SilicaFlickable {
 
         onChatOnlineMemberCountUpdated: {
             if ((chatInformationPage.isSuperGroup || chatInformationPage.isBasicGroup) && chatInformationPage.chatInformation.id.toString() === chatId) {
-                chatInformationPage.chatOnlineMemberCount = chatInformationPage.onlineMemberCount;
+                chatInformationPage.chatOnlineMemberCount = onlineMemberCount;
                 updateGroupStatusText();
             }
         }
--- a/qml/components/chatInformationPage/ChatInformationPageContent.qml
+++ b/qml/components/chatInformationPage/ChatInformationPageContent.qml
@@ -28,7 +28,6 @@ SilicaFlickable {
     property alias membersList: membersList
 
     function initializePage() {
-        console.log("[ChatInformationPage] Initializing chat info page...");
         membersList.clear();
         var chatType = chatInformation.type["@type"];
         switch(chatType) {
@@ -294,7 +293,7 @@ SilicaFlickable {
         id: contentFlickable
         contentHeight: groupInfoItem.height + tabViewLoader.height
         clip: true
-        interactive: true//groupInfoItem.height > pageContent.height * 0.5
+        interactive: !scrollUpAnimation.running && !scrollDownAnimation.running
 
         anchors {
             top: headerItem.bottom
@@ -435,7 +434,7 @@ SilicaFlickable {
                 horizontalAlignment: Qt.AlignHCenter
                 opacity: (tabViewLoader.status === Loader.Ready && tabViewLoader.item.count > 0) ? 1.0 : 0.0
 
-                Behavior on opacity { PropertyAnimation {duration: 500}}
+                Behavior on opacity { FadeAnimation {}}
             }
         }
 
--- a/qml/components/chatInformationPage/ChatInformationPageContent.qml
+++ b/qml/components/chatInformationPage/ChatInformationPageContent.qml
@@ -1,0 +1,459 @@
+/*
+    Copyright (C) 2020 Sebastian J. Wolf and other contributors
+
+    This file is part of Fernschreiber.
+
+    Fernschreiber is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Fernschreiber is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
+*/
+import QtQuick 2.6
+import Sailfish.Silica 1.0
+import "./chatInformationPage"
+import "../js/twemoji.js" as Emoji
+import "../js/functions.js" as Functions
+
+
+SilicaFlickable {
+    id: pageContent
+    property alias membersList: membersList
+
+    function initializePage() {
+        console.log("[ChatInformationPage] Initializing chat info page...");
+        membersList.clear();
+        var chatType = chatInformation.type["@type"];
+        switch(chatType) {
+        case "chatTypePrivate":
+            chatInformationPage.isPrivateChat = true;
+            chatInformationPage.chatPartnerGroupId = chatInformationPage.chatInformation.type.user_id.toString();
+            if(!chatInformationPage.privateChatUserInformation.id) {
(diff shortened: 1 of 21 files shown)


jgibbon:
Whoops, accidentally rebased to monichs nomutex branch, but shouldn't interfere with merging.

qml/pages/ChatPage.qml
```diff
@@ -298,6 +298,8 @@ Page {
         case PageStatus.Active:
             if (!chatPage.isInitialized) {
                 chatModel.initialize(chatInformation);
+
+                pageStack.pushAttached(Qt.resolvedUrl("../pages/ChatInformationPage.qml"), { "chatInformation" : chatInformation, "privateChatUserInformation": chatPartnerInformation, "groupInformation": chatGroupInformation, "chatOnlineMemberCount": chatOnlineMemberCount});
```
monich, line 302: Why "`../pages/"`?

qml/components/chatInformationPage/ChatInformationPageContent.qml
```diff
@@ -0,0 +1,459 @@
...
+import "../js/twemoji.js" as Emoji
+import "../js/functions.js" as Functions
+
+
+SilicaFlickable {
+    id: pageContent
+    property alias membersList: membersList
+
+    function initializePage() {
+        console.log("[ChatInformationPage] Initializing chat info page...");
```
monich, line 31: I suggest to not leave console logs in files committed to master

qml/components/chatInformationPage/ChatInformationPageContent.qml
```diff
@@ -0,0 +1,459 @@
...
+        leftMargin: imageContainer.minDimension + Theme.horizontalPageMargin + Theme.paddingMedium
+        title: chatInformationPage.chatInformation.title !== "" ? Emoji.emojify(chatInformationPage.chatInformation.title, Theme.fontSizeLarge, emojiBase) : qsTr("Unknown")
+        description: chatInformationPage.isPrivateChat ? ("@"+(chatInformationPage.privateChatUserInformation.username || chatInformationPage.chatPartnerGroupId)) : ""
+    }
+
+    SilicaFlickable {
+        id: contentFlickable
+        contentHeight: groupInfoItem.height + tabViewLoader.height
+        clip: true
+        interactive: true//groupInfoItem.height > pageContent.height * 0.5
```
monich, line 297: `interactive` should be false when scroll animation is running, e.g.
```qml
interactive: !scrollUpAnimation.running && !scrollDownAnimation.running
```

qml/components/chatInformationPage/ChatInformationPageContent.qml
```diff
@@ -0,0 +1,459 @@
...
+                text: (chatInformationPage.isPrivateChat && chatInformationPage.privateChatUserInformation.phone_number ? "+"+chatInformationPage.privateChatUserInformation.phone_number : "") || ""
+                isLinkedLabel: true
+            }
+
+            SectionHeader {
+                font.pixelSize: Theme.fontSizeExtraSmall
+                visible: !!inviteLinkItem.text
+                height: visible ? Theme.itemSizeExtraSmall : 0
+                text: qsTr("Invite Link", "header")
+                x: 0
```
monich, line 405: `x` is zero by default

qml/components/chatInformationPage/ChatInformationTabItemBase.qml
```diff
@@ -30,15 +30,12 @@ Item {
...
+    property bool active: index === tabView.currentIndex
 
     default property alias _data: contentItem.data
 
-
-    width: parent.width
-    height: tabView.maxHeight//Math.max(contentItem.height, loadingColumn.height)
-    opacity: active ? 1.0 : 0.0
+    opacity: active ? 1.0 : 0.2
     Behavior on opacity { PropertyAnimation {duration: 300}}
```
monich, line 39: Better to use `FadeAnimation` for all opacity behaviors for visual consistency. Unless this one is made 100ms longer than `FadeAnimation` on purpose (which I doubt)

jgibbon:
Wow, thanks for taking the time to have much more than a look at this. I agree with all of the points raised and will address them ASAP.

qml/components/chatInformationPage/ChatInformationPageContent.qml
jgibbon: For SectionHeader, the default is `x: Theme.horizontalPageMargin` AFAICT

qml/pages/ChatPage.qml
jgibbon: Good point, will change.

qml/components/chatInformationPage/ChatInformationPageContent.qml
jgibbon: Perhaps we should think about log output in general (also c++) in the future. 
Only enable in debug builds? Command line parameter? 
But that's a topic for another place. Will remove here.

qml/components/chatInformationPage/ChatInformationPageContent.qml
jgibbon: Good catch!

qml/components/chatInformationPage/ChatInformationTabItemBase.qml
jgibbon: I actually like 300ms, but you're right. 
Also I missed the 0.2 above which I didn't want to commit.

jgibbon:
I've adressed the mentioned issues (plus a few more) and rebased to current master.

qml/components/chatInformationPage/ChatInformationPageContent.qml
monich: Ah, you're getting rid of margins. OK.

qml/components/chatInformationPage/ChatInformationPageContent.qml
monich: That's one of the reasons why I'm turning those into macros on C++ size - to make it easier to manage those, even strip them completely at compile time.

qml/components/chatInformationPage/ChatInformationTabItemBase.qml
monich: It looks disturbing when different things on the same screen are being animated slightly out of sync))

Wunderfitz:
Works great in most cases. However, a small thing: Please check out channels - it tries to load members of them which results in a neverending spinning wheel. Please try to restore the old behaviour for channels. :)

jgibbon:
@Wunderfitz I think this may be ready now. 
I don't know why github shows conflicts here – my changes should be rebased on master/your merge already.
 
Regarding #170 I've left the header image bigger compared to the ChatPage for now, but the alignment is changed a bit – Silica PageHeader centers the main title to the PageStackIndicator and generally uses a bit more space vertically.

Also, more space is used horizontally for text now when the image is "transitioned down". 
This may be useful for long chat names.

monich:
> Regarding #170 I've left the header image bigger compared to the ChatPage for now, but the alignment is changed a bit – Silica PageHeader centers the main title to the PageStackIndicator and generally uses a bit more space vertically.

And in case if you haven't noticed, `PageHeader` has different height in landscape and portrait.

Wunderfitz:
Yep, looks really good now. If there are other tweaks (e.g. with the page header), we can easily integrate them later. I will merge this one now. Thank you very much!

jgibbon:
> And in case if you haven't noticed, `PageHeader` has different height in landscape and portrait.

That was handled in the initial implementation
`property int minDimension: chatInformationPage.isLandscape ? Theme.itemSizeSmall : Theme.itemSizeMedium`

But you are right, it's an easy thing to overlook, so better be safe than sorry. :)

Issue #154, opened 2020-11-13
ChatPage: Transitions/Restructuring


jgibbon:
At some point we could restructure the ChatPage a bit to allow for additional Items (for now it's mostly the attachment elements but there'll most likely be more to come) to be transitioned in and out for a more fluid UX.

- Perhaps using an anchor based layout may aid with that without having to rely on manual computations in layout properties a lot. 
- Not directly related feature-wise, but it makes sense to do in one go from a development standpoint: Custom properties of some/all child Items could be moved to a QtObject to easily allow using those as Components in Loaders and/or saving a bit of memory.

jgibbon:
I think this issue has become a bit stale and we can get by without changing the ChatPage for the sake of it. 
We can still tackle it if need arises.

PR #169, opened 2020-11-16
Bugfix/message permissions

### LINKED ISSUE #164: new message permissions
- Currently, it's not possible to send messages in a channel. This should be enabled for administrators.
- Message attachment icons should only be visible when permissions are valid.

Both "should" work with (pseudocode) 
`if (privateChat || secretChat || (group && (admin || creator || (member && groupinfo.permissions.can_send_xy))))`

jgibbon:
enhances #159 PR, hopefully fixes #164 issue
### Info for "beta testers" ###
If anyone uses groups with message restrictions (you're not allowed to post messages, videos, polls, stickers, …), a bit of help testing would be very appreciated. 
This applies to the whole message entry area, the attachment icon buttons (should be hidden) as well as forwarding messages (chats should become hidden; you can forward by selecting messages via long press context menu). Thanks in advance! ;)

### ChatPage ###
**Please mind,** I wasn't able to verify all permissions in real world examples, so a short sanity check of changes is strongly advised to make sure I didn't miss anything serious.

According to TDLib documentation, animations and videos use different permissions for some reason.
The attachment icon is only dependent on "can_send_media_messages" (video), because AFAICT we only send those.

I've not yet set anything up for secret chats (despite "proposing" it in #164) because that heavily depends on how those will be implemented.

This PR has a slight performance impact on the ChatPage, but that should sort itself out with the next round of optimizations (perhaps while implementing #154).

### ChatSelectionPage ###
Changes try to hide chats where the selected messages can't be forwarded to. 
This is very expensive computationally, since it has to check the member status (inside groupInformation, which has to be aquired) for every group chat.
For that reason, initially, all delegates are shown dimmed (but not clickable) and the permission check has been off-loaded to an asynchronous Loader. 
The "delegate-based filtering" has some draw-backs (like showing/hiding when scrolling back and forth), but it gets stuff drawn bearably fast on initialization.

I tried to keep the ChatSelectionPage easily extendable for future purposes (perhaps system integration for sharing texts/urls/media?) by using/checking its "state" property. Right now, it only uses the string "forwardMessages".

cheers

### DIFF
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -412,7 +412,7 @@ function getMessagesNeededForwardPermissions(messages) {
 
     var mediaMessageTypes = ["messageAudio", "messageDocument", "messagePhoto", "messageVideo", "messageVideoNote", "messageVoiceNote"];
     var otherMessageTypes = ["messageAnimation", "messageGame", "messageSticker"]
-    for(var i = 0; i < messages.length; i += 1) {
+    for(var i = 0; i < messages.length && neededPermissions.length < 3; i += 1) {
         var type = messages[i]["content"]["@type"];
         var permission = "";
         if(type === "messageText") {
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -406,3 +406,28 @@ function handleErrorMessage(code, message) {
         appNotification.show(message);
     }
 }
+
+function getMessagesNeededForwardPermissions(messages) {
+    var neededPermissions = ["can_send_messages"];
+
+    var mediaMessageTypes = ["messageAudio", "messageDocument", "messagePhoto", "messageVideo", "messageVideoNote", "messageVoiceNote"];
+    var otherMessageTypes = ["messageAnimation", "messageGame", "messageSticker"]
+    for(var i = 0; i < messages.length; i += 1) {
+        var type = messages[i]["content"]["@type"];
+        var permission = "";
+        if(type === "messageText") {
+            continue;
+        } else if(type === "messagePoll") {
+            permission = "can_send_polls";
+        } else if(mediaMessageTypes.indexOf(type) > -1) {
+            permission = "can_send_media_messages";
+        } else if(otherMessageTypes.indexOf(type) > -1) {
+            permission = "can_send_other_messages";
+        }
+
+        if(permission !== "" && neededPermissions.indexOf(permission) === -1) {
+            neededPermissions.push(permission);
+        }
+    }
+    return neededPermissions;
+}
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -408,26 +408,26 @@ function handleErrorMessage(code, message) {
 }
 
 function getMessagesNeededForwardPermissions(messages) {
-    var neededPermissions = ["can_send_messages"];
+    var neededPermissions = ["can_send_messages"]
 
-    var mediaMessageTypes = ["messageAudio", "messageDocument", "messagePhoto", "messageVideo", "messageVideoNote", "messageVoiceNote"];
+    var mediaMessageTypes = ["messageAudio", "messageDocument", "messagePhoto", "messageVideo", "messageVideoNote", "messageVoiceNote"]
     var otherMessageTypes = ["messageAnimation", "messageGame", "messageSticker"]
     for(var i = 0; i < messages.length && neededPermissions.length < 3; i += 1) {
-        var type = messages[i]["content"]["@type"];
-        var permission = "";
+        var type = messages[i]["content"]["@type"]
+        var permission = ""
         if(type === "messageText") {
-            continue;
+            continue
         } else if(type === "messagePoll") {
-            permission = "can_send_polls";
+            permission = "can_send_polls"
         } else if(mediaMessageTypes.indexOf(type) > -1) {
-            permission = "can_send_media_messages";
+            permission = "can_send_media_messages"
         } else if(otherMessageTypes.indexOf(type) > -1) {
-            permission = "can_send_other_messages";
+            permission = "can_send_other_messages"
         }
 
         if(permission !== "" && neededPermissions.indexOf(permission) === -1) {
-            neededPermissions.push(permission);
+            neededPermissions.push(permission)
         }
     }
-    return neededPermissions;
+    return neededPermissions
 }
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -270,8 +270,8 @@ Page {
         forwardMessagesTimer.start();
     }
     function hasSendPrivilege(privilege) {
-        var groupStatus = chatGroupInformation ? chatGroupInformation.status : null;
-        var groupStatusType = groupStatus ? groupStatus["@type"] : null;
+        var groupStatus = chatGroupInformation ? chatGroupInformation.status : null
+        var groupStatusType = groupStatus ? groupStatus["@type"] : null
         return chatPage.isPrivateChat
                     || (groupStatusType === "chatMemberStatusMember" && chatInformation.permissions[privilege])
                     || groupStatusType === "chatMemberStatusAdministrator"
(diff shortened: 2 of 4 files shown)


qml/js/functions.js
```diff
@@ -406,3 +406,28 @@ function handleErrorMessage(code, message) {
...
+        } else if(otherMessageTypes.indexOf(type) > -1) {
+            permission = "can_send_other_messages";
+        }
+
+        if(permission !== "" && neededPermissions.indexOf(permission) === -1) {
+            neededPermissions.push(permission);
+        }
+    }
+    return neededPermissions;
+}
```
monich, line 433: This function looks like a perfect candidate for being implemented in C++. Moving QVariantMaps between C++ and QML is A LOT cheaper than this sort of JavaScript code.

qml/pages/ChatPage.qml
```diff
@@ -266,6 +266,17 @@ Page {
...
         forwardMessagesTimer.start();
     }
+    function hasSendPrivilege(privilege) {
+        return chatPage.isPrivateChat ||
+                chatGroupInformation &&
+                (
+                    (chatGroupInformation.status["@type"] === "chatMemberStatusMember" && chatGroupInformation.status.permissions[privilege])
+                    || chatGroupInformation.status["@type"] === "chatMemberStatusAdministrator"
+                    || chatGroupInformation.status["@type"] === "chatMemberStatusCreator"
+                    || (chatGroupInformation.status["@type"] === "chatMemberStatusRestricted" && chatGroupInformation.status.permissions[privilege])
```
monich, line 276: Each `chatGroupInformation.status["@type"]` is a lookup in QVariantMap (too lookups actually, `chatGroupInformation.value("status").toMap().value("@type")`), each returning the same result. And being done from QML, it's significantly slower than its C++ equivalent. Better to calculate it once and then compare against constants. Similarly, `chatGroupInformation.status.permissions[privilege]` is 3 lookups, done twice. Even better would be to expose corresponding properties from the model and compare them against enums, at least the `@type` thing...

qml/pages/ChatSelectionPage.qml
```diff
@@ -61,8 +67,53 @@ Dialog {
...
+            Loader { // checking permissions takes a while, so we defer those calculations
+                id: visibleLoader
+                asynchronous: true
+                sourceComponent: Component {
+                    QtObject {
+                        property var chatGroupInformation: ({})
+                        property bool visible: false
+                        Component.onCompleted: {
+                            if(chatSelectionPage.state === "forwardMessages") {
+                                var chatType = display.type['@type'];
```
monich, line 80: Here we go, you do optimize QVariantMap lookups sometimes... :)

qml/pages/ChatSelectionPage.qml
```diff
@@ -61,8 +67,53 @@ Dialog {
...
+                                else if (chatType === "chatTypeBasicGroup" ) {
+                                    chatGroupInformation = tdLibWrapper.getBasicGroup(display.type.basic_group_id);
+                                }
+                                else if (chatType === "chatTypeSupergroup" ) {
+                                    chatGroupInformation = tdLibWrapper.getSuperGroup(display.type.supergroup_id);
+                                }
+
+                                visible = (chatGroupInformation.status["@type"] === "chatMemberStatusCreator"
+                                        || chatGroupInformation.status["@type"] === "chatMemberStatusAdministrator"
+                                        || (chatGroupInformation.status["@type"] === "chatMemberStatusMember"
```
monich, line 94: but not always :/

monich:
Use of semicolons in JavaScript is pretty inconsistent. There's no obvious logic why they are being used here and not being used there. I suggest to drop those which are not necessary (which means most of them)

jgibbon:
Thank you for another really good review! I'll check your comments ASAP and try (not very confident in my c++ skills as you probably can tell) to integrate your suggestions.

qml/js/functions.js
jgibbon: Sometimes I almost forget that we're dealing with a really old and comparatively slow js engine here. ;) 
I'll try my best.

qml/pages/ChatPage.qml
jgibbon: Again, it's easy to forget we're not dealing with array lookups inside a current JS engine (which usually are amazingly fast depending on scope lookups) – thanks for spelling it out for me!

qml/pages/ChatSelectionPage.qml
jgibbon: In this case, I can actually omit the property completely (isn't used outside the function at all) and just use a JS var. 
But there's another bug hidden in there – thanks for making me aware.

qml/js/functions.js
jgibbon: @monich After implementing a (really simple) C++ alternative, I discovered that this computation really isn't very relevant: 

For 50 messages on my Xperia X, JS gets it done at roughly 0-1ms*/150µs**, while my admittedly pretty unoptimized c++ takes (including communication overhead) roughly 5ms* for the same messages. 

I consider forwarding > 50 messages at the same time an edge case.

Here's my ultra basic c++ for reference, even I can guess one could preallocate strings or even lists.
But with that JS performance, I don't think it's really worth the effort regardless of further c++-optimizations, at least in this case. Agreed?

```
QStringList FernschreiberUtils::getMessagesNeededForwardPermissions(const QVariantList &messages)
{
    QStringList permissions({"can_send_messages"});
    QStringList mediaMessageTypes({"messageAudio", "messageDocument", "messagePhoto", "messageVideo", "messageVideoNote", "messageVoiceNote"});
    QStringList otherMessageTypes({"messageAnimation", "messageGame", "messageSticker"});

    QString permission;

    for (QVariantList::const_iterator j = messages.begin(); j != messages.end() && permissions.count() < 3; j++)
    {
        QString messageType = j->toMap().value("content").toMap().value("@type").toString();
        if(messageType == "messageText") {
            continue;
        }
        else if(messageType == "messagePoll") {
            permission = "can_send_polls";
        }
        else if(mediaMessageTypes.contains(messageType)) {
            permission = "can_send_media_messages";
        }
        else if(otherMessageTypes.contains(messageType)) {
            permission = "can_send_other_messages";
        }
        if(!permissions.contains(permission)) {
            permissions.append(permission);
        }
    }
    return permissions;
}

```

*) Really simple Date comparison from JS, so only ms granularity.
**) QML Profiler

monich: Hmm... I would have expected some noticeable difference. Regarding your C++ code, commonly used `QStrings` can be statically or locally pre-allocated (or `QStringLiteral` be used) to avoid ASCII => UCS2 conversions of `char*`, but I doubt that it would be a huge saver. I guess `QVariantMap` is so inefficient that once you start using it, you're doomed in terms of performance))

If message was a pre-parsed struct/class, its type a enum and so on, I'm sure C++ would perform much better.

Oh well. I would still prefer C++ because it's easier to debug, but if you're more comfortable with JavaScript, then go ahead, of course.

qml/pages/ChatPage.qml
jgibbon: done

jgibbon:
The mentioned issues should now be adressed.

> Use of semicolons in JavaScript is pretty inconsistent. There's no obvious logic why they are being used here and not being used there. I suggest to drop those which are not necessary (which means most of them)

You are right, there was some inconsistency. 
Generally I am in favor of keeping semicolons (at least in "pure" JS), primarily for the following reasons:
 - It spares the parser from "adding them internally"
 - There may be cases where expected behaviour differs from actual behaviour ([examples](https://flaviocopes.com/javascript-automatic-semicolon-insertion/))
 - Not a "good" reason, but an opinion: There are a lot of cases where it's easier to "parse by eye". For example, you see at first glance if that "}" ends a code block or an object definition.

Of course, (inline) JS inside QML is a bit different and I get that many QML folks prefer to reduce them as much as possible.

**I've now removed most semicolons** in the parts changed by this PR (creating even more inconsistencies to be honest, but I didn't want to change unrelated parts too much in this context).

Thanks again for your review, it was very much appreciated!

monich:
> Not a "good" reason, but an opinion: There are a lot of cases where it's easier to "parse by eye". For example, you see at first glance if that "}" ends a code block or an object definition.

It's subjective, I know. But some things are more important than the others, different things for different people, you never know before you ask. The overall approach can be sort of "optimized" to minimize the total amount of pain inflicted on those involved. And I would say, opinions of those who contribute more (like you), should be given more weight.

To me, end of line is a good separator by itself and unnecessary semicolons is visual garbage which distracts and actually makes it harder to comprehend the text. Like tons of `this->` in C++ code from @Wunderfitz )) I just can't force myself to add those. It feels like, instead of writing a simple sentence you have to insert a bunch of unnecessary blah-blah-blah between the words. Of blah course blah it blah can blah be blah readable blah when blah you blah get blah used blah to blah it blah but blah meh :)

And to me it's also sort of related to the principle "don't introduce entities without necessity"

Wunderfitz:
Yep, couldn't agree more to @monich. Things are highly subjective and you can probably have lengthy discussions about code style (not mentioning tabs vs. spaces) without writing a single line of code. ;) I've also become much more relaxed in that way over the last few weeks compared to my statements when the first contributions especially by @monich came in. The main thing is that we get along somehow and can continue working together. 

If you ask me personally, I'm a big fan of semicolons in JS code (as you can probably easily grasp from my code), but as said, I've become much more relaxed. ;) And concerning the usage of `this->` - I simply want to see immediately if I'm assigning/using an attribute or a simple variable. Of course, you can also handle that by convention (by using underscores for instance), but well, using `this->` is something I feel strangely at home with (might be related that I used to code a lot of Java). 

Back to topic: From my perspective, the proposed change works smoothly and I would merge it now to have a solid foundation for other changes that might come (still need to do some changes to the pin message thingy...). Thanks again!

PR #177, opened 2020-11-18
Feature: Support pinned messages

### LINKED ISSUE #57: Notification - Pinned Messages
Hi,

thank you so much for Fernschreiber1

I encountered the issue that when I muted chats I get a notification that the pinned message was updated anyways. Also in the notification and in the chat it is written: "Unsupported message: PinMessage". 

It is kind of annoying. How can I disable that? I do not find any option in the pully menu. 

Is this a bug?

Wunderfitz:
This is the replacement for the ill-fated PR #175 which I completely messed up. It should include all the work and the additional changes which were proposed by @jgibbon. Moreover, unpinning messages is now supported as well. Happy testing!

This will fix #57 completely (was already partially done before)


qml/components/MessageListViewItem.qml
```diff
@@ -277,11 +284,22 @@ ListItem {
...
+                                inReplyToMessage: messageInReplyToLoader.inReplyToMessage
+                            }
+                            MouseArea {
+                                anchors.fill: parent
+                                onClicked: {
+                                    messageOverlayLoader.overlayMessage = messageInReplyToRow.inReplyToMessage;
+                                    messageOverlayLoader.active = true;
+                                }
+                            }
                         }
```
jgibbon, line 303: Perhaps you don't even need the Item at all…
```suggestion
                        InReplyToRow {
                            id: messageInReplyToRow
                            myUserId: page.myUserId
                            visible: true
                            inReplyToMessage: messageInReplyToLoader.inReplyToMessage
                            
                            MouseArea {
                                anchors.fill: parent
                                onClicked: {
                                    messageOverlayLoader.overlayMessage = messageInReplyToRow.inReplyToMessage;
                                    messageOverlayLoader.active = true;
                                }
                            }
                        }
```

jgibbon (changes requested):
Still works fine for general usage, but there's an issue handling the unpin. 

I ran into the Telegram rate limit (that was a first!) while testing, so here it is from memory, a bit rough:
- Pin two messages, unpin one, the displayed pinned message does not update.
- (could be because of rate limit? but I did get the error only afterwards) At one point after pinning a message, the area was displayed but empty.

Other than that: Really nice feature!

Wunderfitz:
I also had the rate limiting (too many pins/unpins triggers it, yes...). But I couldn't reproduce the issue that the area is displayed empty. Strange... If you have any more concrete hint, please let me know. I'll check!

Concerning the Item - I also thought I won't need it. However, afterwards, it looked like that:

image

So for now, I'll keep it that way... ;)

Wunderfitz:
Thx, then let's merge it!

PR #179, opened 2020-11-19
Bugfix/#20 translation plurals

### LINKED ISSUE #20: Plurals in translations
Hi,

Could you, please, use special formatting for plurals? For example

```diff
- %1 members
+ %n members
```

In some languages (for example, in Russian) there are several forms of plural. Formatting with `%1` looks really ugly.

UPD1: Also this string should be in plural `unread messages`. Something like

```qml
qsTr("unread message(s)", "0", count)
```

jgibbon:
This should fix #20.

I've also added  `language="en"` to the base `harbour-fernschreiber.ts` – while it doesn't seem to be used for in-app output (it shows the same "untranslated" output when run from SDK, for example), this seems to generate/retain the `<numerusform>`-tags, which should be less confusing for translators using it as a base.

Sadly, it's inevitable that the "broken" strings now have to be re-translated.

Funny anecdote: 
At some point I really scratched my head, because I didn't think about how number placeholders like `%Ln` handle replacements within qsTr(), while `%i` requires a dedicated `.arg()` call. So I had a lot of useless `.arg()` calls spewing warnings left and right. While that took me more time I care to admit, these are removed now and it seems to work nicely.

Cheers!

### DIFF
--- a/qml/components/PollPreview.qml
+++ b/qml/components/PollPreview.qml
@@ -178,7 +178,7 @@ Item {
                     Label {
                         id: optionVoterPercentage
                         font.pixelSize: Theme.fontSizeTiny
-                        text: qsTr("%L1\%", "% of votes for option").arg(modelData.vote_percentage)
+                        text: qsTr("%Ln\%", "% of votes for option").arg(modelData.vote_percentage)
                         horizontalAlignment: Text.AlignRight
                         anchors {
                             right: parent.right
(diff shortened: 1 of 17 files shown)


Wunderfitz:
Thanks, looking good as far as I could see and test. Yep, the downside is that all translations need to be adjusted, but at least it seems that the old translations are at least copied over to a certain extent. So, let's merge it. 

Heads-up to the translators: @dashinfantry @jorm1s @edp17 @iamnomeutente @atlochowski @arustg @GNUuser @eson57

jgibbon:
I managed to change some manually, too, to save the translators a bit of work, at least. :)

Wunderfitz:
:+1:

PR #184, opened 2020-11-20
Show notification popups except for the current chat

monich:
And a few other related tweaks:

1. Drop dependency on Ngf
2. Pre-initialize the chat photo when pushing the chat page.

The "display on" option for popups will be made configurable once this is merged.


Wunderfitz (approved):
Works smoothly here, thanks for the changes! Impressed about the undocumented (?) hints for Nemo Notifications... If I had known them earlier...

PR #185, opened 2020-11-20
Add switchable Debug output

### LINKED ISSUE #174: [discussion] handling debug output
This isn't really high priority for now, but I'd like to hear your thoughts on handling debug output, because at some point before 1.0 we should adress this IMHO.

- Should it only be available in debug builds?
- Should it be accessible via command line parameter?
- (How) do we handle QML output (console)?

The first option would result in a "cleaner" release build, whereas the second could aid users helping out with bug hunting. 

Basically the same goes for QML – sometimes it's useful, but it's mostly good for debugging, which most users won't ever do. Also I'm not aware of a (sane) way to get rid of QML/JS parts at compile time. Handling it at runtime would mean a bit of performance penalty even for "not using" it.

What do you think?

cheers!

jgibbon:
This PR aims to fix #174 when it's ready, but until then is a place to work on those changes together. 

As announced in the issue, I've prepared the QML/JS part based on the ideas discussed there as a start.

The JS part is basically a drop-in replacement for `console`, so it covers
 - `console.log`
 - `console.assert`
 - `console.time` & - `console.timeEnd`
 - `console.trace`
 - `console.count`
 - `console.profile`
 - `console.exception`

If you previously would've used `console.*`, make sure to:
 - import `debug.js`
 - just use `Debug.*` instead

Also, I've changed functions.js to a shared JavaScript resource (".pragma library") to make usage in there easier. According to the qml docs, this also "can save significant amounts of memory and speed up the instantiation of QML components". Win-win.

For now, the C++ part is basically copied from the issue comments by @monich with a small `ifdef` thrown in.

### DIFF
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -21,6 +21,7 @@ import Sailfish.Silica 1.0
 import "../js/twemoji.js" as Emoji
 import "../js/functions.js" as Functions
 import QtQml.Models 2.3
+import "../js/debug.js" as Debug
 
 ListItem {
     id: messageListItem
@@ -136,12 +137,12 @@ ListItem {
             messageBackground.isUnread = index > chatModel.getLastReadMessageIndex();
         }
         onLastReadSentMessageUpdated: {
-            console.log("[ChatModel] Messages in this chat were read, new last read: " + lastReadSentIndex + ", updating description for index " + index + ", status: " + (index <= lastReadSentIndex));
+            Debug.log("[ChatModel] Messages in this chat were read, new last read: ", lastReadSentIndex, ", updating description for index ", index, ", status: ", (index <= lastReadSentIndex));
             messageDateText.text = getMessageStatusText(myMessage, index, lastReadSentIndex, messageDateText.useElapsed);
         }
         onMessageUpdated: {
             if (index === modelIndex) {
-                console.log("[ChatModel] This message was updated, index " + index + ", updating content...");
+                Debug.log("[ChatModel] This message was updated, index ", index, ", updating content...");
                 messageDateText.text = getMessageStatusText(myMessage, index, chatView.lastReadSentIndex, messageDateText.useElapsed);
                 messageText.text = Emoji.emojify(Functions.getMessageText(myMessage, false, messageListItem.isOwnMessage), messageText.font.pixelSize);
             }
--- a/src/debuglogjs.h
+++ b/src/debuglogjs.h
@@ -1,0 +1,52 @@
+/*
+    Copyright (C) 2020 Sebastian J. Wolf and other contributors
+
+    This file is part of Fernschreiber.
+
+    Fernschreiber is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Fernschreiber is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef DEBUGLOG_H
+#define DEBUGLOG_H
+
+
+#include <QObject>
+#include <QQmlEngine>
+
+class DebugLogJS : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(bool enabled READ isEnabled WRITE setEnabled NOTIFY enabledChanged)
+public:
+
+#ifdef QT_QML_DEBUG
+    DebugLogJS(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(true) {}
+#else
+    DebugLog(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(false) {}
+#endif
+    static QObject* createSingleton(QQmlEngine*, QJSEngine*) { return new DebugLogJS(); }
+    bool isEnabled() const { return enabled; }
+    void setEnabled(bool value) {
+        if (enabled != value) {
+            enabled = value;
+            Q_EMIT enabledChanged(value);
+        }
+    }
+Q_SIGNALS:
+    void enabledChanged(bool value);
+private:
+    bool enabled;
+};
+
+#endif // DEBUGLOG_H
--- a/src/debuglogjs.h
+++ b/src/debuglogjs.h
@@ -23,6 +23,7 @@
 
 #include <QObject>
 #include <QQmlEngine>
+#include <QLoggingCategory>
 
 class DebugLogJS : public QObject
 {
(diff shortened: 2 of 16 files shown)


qml/components/MessageListViewItem.qml
```diff
@@ -136,12 +137,12 @@ ListItem {
             messageBackground.isUnread = index > chatModel.getLastReadMessageIndex();
         }
         onLastReadSentMessageUpdated: {
-            console.log("[ChatModel] Messages in this chat were read, new last read: " + lastReadSentIndex + ", updating description for index " + index + ", status: " + (index <= lastReadSentIndex));
+            Debug.log("[ChatModel] Messages in this chat were read, new last read: " + lastReadSentIndex + ", updating description for index " + index + ", status: " + (index <= lastReadSentIndex));
```
monich, line 140: Would this work too?
```JavaScript
Debug.log("[ChatModel] Messages in this chat were read, new last read:", lastReadSentIndex, ", updating description for index", index, ", status:", (index <= lastReadSentIndex));
```
That would minimize runtime penalty to a minimum by letting `console.log()` (if it ends up being invoked) to do the concatenation.

qml/components/MessageListViewItem.qml
```diff
@@ -136,12 +137,12 @@ ListItem {
...
             messageDateText.text = getMessageStatusText(myMessage, index, lastReadSentIndex, messageDateText.useElapsed);
         }
         onMessageUpdated: {
             if (index === modelIndex) {
-                console.log("[ChatModel] This message was updated, index " + index + ", updating content...");
+                Debug.log("[ChatModel] This message was updated, index " + index + ", updating content...");
```
monich, line 145: Here too... and there are many more places where strings are being allocated and formatted only to be thrown away if debug output is disabled.

qml/components/MessageListViewItem.qml
jgibbon: Good Idea & done.

src/debuglog.h
```diff
@@ -0,0 +1,52 @@
...
+    Q_OBJECT
+    Q_PROPERTY(bool enabled READ isEnabled WRITE setEnabled NOTIFY enabledChanged)
+public:
+
+#ifdef QT_QML_DEBUG
+    DebugLog(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(true) {}
+#else
+    DebugLog(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(false) {}
+#endif
+    static QObject* createSingleton(QQmlEngine* aEngine, QJSEngine* aScript) { return new DebugLog(); }
```
monich, line 38: Drop these `aEngine`  and `aScript` things, they are unused (and look weird next to everything else)

src/debuglog.h
```diff
@@ -0,0 +1,52 @@
...
+class DebugLog : public QObject
+{
+    Q_OBJECT
+    Q_PROPERTY(bool enabled READ isEnabled WRITE setEnabled NOTIFY enabledChanged)
+public:
+
+#ifdef QT_QML_DEBUG
+    DebugLog(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(true) {}
+#else
+    DebugLog(QObject* parent = Q_NULLPTR) : QObject(parent), enabled(false) {}
```
monich, line 36: How about initializing `enabled` from an environment variable?

src/debuglog.h
jgibbon: When I removed those, It didn't build, because apparently `qmlRegisterSingletonType` expected them…

src/debuglog.h
jgibbon: Yeah, I wasn't sure how you wanted to do the "real" c++ part and left that one as basic as possible as a result.

I thought we could use `QCommandLineParser` to introduce a `-v` argument for non-debug builds, possibly in addition to environment variables? What do you suggest?

Sorry for the noob question, but I guess, the `static` declaration would be enough to ensure it's the same instance if I called `createSingleton` from `harbour-fernschreiber.cpp` again, right?

src/debuglog.h
monich: I mean like this:
```C++
static QObject* createSingleton(QQmlEngine*, QJSEngine*) { return new DebugLog(); }
````

File: src/debuglog.h (hunk 4 above)
Comment (monich) [reply src/debuglog.h:36]: `DebugLog::createSingleton` (or whatever is passed to `qmlRegisterSingletonType<DebugLog>`) is called by QML engine when it needs to access `DebugLog` object for the first time. Lazy instantiation on demand. QML engine owns this object and will destroy it at shutdown. There's no guarantee when (if ever) this singleton will be instantiated. Which makes it a bit inconvenient to interact with this instance from C++ side.

The `enabled` field can be initialized from a global variable (which in turn can be initialized in `main` before even registering the singleton). Alternatively or in addition to that, `DebugLog::createSingleton` can set a global variable pointing to this singleton instance instantiated and owned by QML engine, the destructor can set it to NULL (checking first that it's equal to `this`, just in case) if it needs to be accessed from C++ later in the game (which is probably unnecessary). Something along these lines can be done.

Comment (monich) [reply src/debuglog.h:36]: Or you can just leave it as is (leaving it enabled by default) and I can integrate that later with logging categories.

Comment (jgibbon) [reply src/debuglog.h:36]: Yeah, maybe that would get us a better result ;)

File: src/debuglog.h (hunk 3 above)
Comment (jgibbon) [reply src/debuglog.h:38]: Oh, thanks!

**Comment by jgibbon** [comment]
@monich I've now updated/rebased/force-pushed the changes for using QLoggingCategory in the front end, as well.

I've also written some stuff in the README, 
including mentioning some of our contributions that came to mind. 

It worked in my tests, but I'm sure you have some ideas to improve this – feel free to directly push to this branch. Thanks!

File: src/debuglogjs.h [hunk 5]
```diff
@@ -0,0 +1,52 @@
...
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef DEBUGLOG_H
+#define DEBUGLOG_H
+
+
```
monich, line 23: Empty lines randomly appearing and disappearing here and there don't help to read the code :)

jgibbon: Oh, also I haven't changed the DEBUGLOG_H thing! Ooops!

src/debuglogjs.h
```diff
@@ -0,0 +1,52 @@
...
+    static QObject* createSingleton(QQmlEngine*, QJSEngine*) { return new DebugLogJS(); }
+    bool isEnabled() const { return enabled; }
+    void setEnabled(bool value) {
+        if (enabled != value) {
+            enabled = value;
+            Q_EMIT enabledChanged(value);
+        }
+    }
+Q_SIGNALS:
+    void enabledChanged(bool value);
```
monich, line 46: There's no need to pass the new value with the signal. I think in any case QML runtime will call the getter, but it's not only that. It's not particularly useful and under certain circumstances may even be wrong. Consider the case when one of the signal handlers (not the last one) changes the value again. I'm not sure how that would be handled - most likely, the remaining handlers will receive the stale value as a parameter. To avoid uncertainties like that, better to not pass anything. Whoever wants to know the new value, will call the getter and get the actual latest value.

src/debuglogjs.h
```diff
@@ -0,0 +1,52 @@
...
+    Fernschreiber is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#ifndef DEBUGLOG_H
```
monich, line 20: Make is `DEBUGLOGJS_H` or something. I used `FERNSCHREIBER_DEBUG_LOG_H` in `debuglog.h` because `debuglog` is such a generic name that it's easy to imagine `#ifndef DEBUGLOG_H` appearing in one of the headers indirectly pulled in from elsewhere.

monich:
A few nitpicks, overall LGTM!

jgibbon:
OK, I'll just address those quickly. Thank you!

PR #191, opened 2020-11-22
Take logging categories into use

monich:
Logging can be enabled in release build like this:
```bash
QT_LOGGING_RULES="fernschreiber.*=true" harbour-fernschreiber
```
See https://doc.qt.io/qt-5/qloggingcategory.html for more details.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -145,6 +145,7 @@ HEADERS += \
     src/chatmodel.h \
     src/dbusadaptor.h \
     src/dbusinterface.h \
+    src/debuglog.h \
     src/emojisearchworker.h \
     src/fernschreiberutils.h \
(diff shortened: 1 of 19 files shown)


jgibbon:
Oh, a slight misunderstanding: I thought you'd just take over the other PR/branch and add to that. 
But no worries, the only thing that collides is the `debuglog.h` but that may be easily merged..

I guess it would now make sense to get this one merged first and then rebase #185 to this to continue there?

cheers

monich:
Oh, I haven't actually noticed `debuglog.h`, my bad :/

monich:
Maybe rename it into something like `debuglogjs.h` or something? This one doesn't need to be included from every C++ file.

jgibbon:
Yeah, or *frontend.h or something. Makes sense.

PR #192, opened 2020-11-22
Emoji tweaks

jgibbon:
As hinted in the Telegram channel, I have stripped the twemoji.js to the bare minimum we're using to reduce a bit of overhead.

- `twemoji.js` was stripped down a lot (588 lines → 66 lines) and made a "`.pragma library`" to reduce initialization cost (both memory and computation) for every Item where it is imported from. The performance benefits for calling "`emojify`" itself are big for strings with unrealisticly large amounts of emojis (~16% with many thousands), but negligible for normal usage. Nonetheless, this stripped down version may even help figuring out how to possibly implement it in c++ at some point, since there's not so much "unused rubbish" to wade through.
- `basePath` is now defined once within twemoji.js. It uses `Qt.resolvedUrl` to get an absolute path which works from everywhere. So the old workarounds for that became obsolete and were removed.
- I've tried to identify all places where `Text`s where elided and changed that to `Label`/`TruncationMode.Fade`. For the `ChatPage` header, that meant a few other changes, since the Truncation direction changes with horizontalAlignment. Switching to `Label` also meant I could remove some `color` declarations, since it defaults to `primaryColor`.
- Also, I've updated the Twemoji image assets. A few are a bit smaller, while others use a more diversified approach not focussing on "white, blond, blue eyes" as far as I could tell. There are no new images in the current twemoji version, only updated ones, which made this pretty easy.


Wunderfitz:
Thanks for all the work. As said in the Telegram chat, I was sceptical, but it really looks nice. So let's merge it. :)

PR #193, opened 2020-11-22
Add MceInterface object

monich:
Combines all mce related functionality in one place, reuses `QDBusInterface`.


Wunderfitz:
Looking good, thanks!

PR #199, opened 2020-11-23
fix functions scope; make appNotification global

jgibbon:
appNotification now resides (only) in the appWindow scope, where it is accessible from everywhere.
Defining the parent takes care of sizing/rotation.
I've left some instances of AppNotification items where they're supposed to be centered inside a component, not the whole page.

functions.js globals (tdLibWrapper/appNotification) get set in harbour-fernschreiber.qml, as well.


Wunderfitz:
Thx, that makes links and other things work again, thanks! However, still puzzled why it doesn't work if I have QML live bench turned on...

PR #200, opened 2020-11-23
Added storage optimizer option

monich:
I'm not quite sure what it does but it sounds important :)

### DIFF
--- a/qml/pages/SettingsPage.qml
+++ b/qml/pages/SettingsPage.qml
@@ -149,6 +149,23 @@ Page {
                 }
             }
 
+            SectionHeader {
+                text: qsTr("Storage")
+            }
+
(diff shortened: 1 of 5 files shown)


Wunderfitz:
No problem with merging unfinished stuff, but as long as the option doesn't do anything, maybe we should hide it from the default settings view, shouldn't we?

monich:
It's supposed to periodically [remove](https://github.com/tdlib/td/issues/667) old data.

Wunderfitz:
That's it? Well, OK - then let's merge it! :D

monich:
If it actually works (which it probably does) then later its parameters (size limit, time limit etc.) can be made configurable too.

PR #203, opened 2020-11-26
Optimize functions.js

jgibbon:
## JS Optimizations:
Generally speaking, performance wasn't bad.
I didn't change the "execution flow" too much for the most part, just optimized a few things here and there. This managed to get the memory allocations down for almost all of the changed functions mentioned below (only exception: getUserName stayed the same) and – at least with my limited test data – execution time also was reduced in most runs.

For roughly getting an idea of how the changed functions perform, the new `Functions.compareAndRepeat` was used repeatedly (multiple Application launches and/or at least visits to the `AboutPage`). 
To simplify checking these changes without too much interference, the test runs (and import of the unchanged `functions.js`) are left on the `AboutPage`. Please feel free to add to them if you think of additional edge cases.

**This must be either removed or stripped & commented out** before converting this draft PR to a final one.

To get an idea how to use this, check the following video I made:
https://youtu.be/HHnflwbKqAE
The video was made using the emulator for presentation purposes only. 
  - Testing on a device is, of course, recommended. It is (quite a lot) slower, don't mind the "application not responding" layer and please wait for the results patiently. It can take a seemingly unresponsive while to stop profiling.
  - You may want to go offline to prevent loads of JSON output when you want to see the result comparison in the "Application Output" pane of the SDK.

#### `compareAndRepeat`:
This new function (**now in debug!**) is only used to compare new implementations with old ones and is left in to simplify tests like these in the future.
After trying and discarding a few approaches to measure/compare execution time inside of JS, I came to the (almost obvious) conclusion to concentrate on comparing return values + repeating and to leave measurements to the QML Profiler. This way we get to see memory consumption inside the functions, as well.

A description of arguments is added inside the JS, but here's a hypothetical example (this would get a lot of console output):
```JavaScript
/*
To test
Debug.log("first", "second", "third")
and
Debug.log("some value")
against the same calls in console.log 100 times:
*/
Debug.compareAndRepeat("Debuglog", Debug.log, console.log, [["first", "second", "third"], ["some value"]], 100);
```

#### `getUserName`: 
Basically just removed the used variables. This may be the most insignificant change.

#### `getMessageText`: 
Turned to a switch/case. It's a bit faster and uses less memory.

####  `handleHtmlEntity`:
Removed some operations by handling comparison + setting nextIndex "in one go". Also used the simplified message insertion (see below).

####  `enhanceHtmlEntities`:
The messageInsertions logic was replaced by a few replace calls. This uses only a fraction of the memory and computation time.

####  `enhanceMessageText`:
In JS, `new Something()` basically equals initializing a `Class` (current JS also has "real" Classes, but that's not much more than syntactic sugar), which normally is used for "persistent logic blocks".
By replacing this with a simple object notation, it's not necessary to invoke the function and prepare the scope including custom prototype. This is a bit cheaper for memory/computation.

Also, the "ifs" were replaced with a switch/case and creation of regular expression and sorting function were moved outside of `enhanceMessageText`.
Since `Array.push` supports multiple arguments, I made use of that feature to remove some calls.
Before the two `for` loops, there is a check if they are necessary. Otherwise, `String.replace` (like in `enhanceHtmlEntities`) is used.


jgibbon:
After talking with @Wunderfitz, 
I went ahead and implemented his useful suggestions:

 - move `compareAndRepeat` inside `debug.js`
 - create a new `DebugPage` for tests instead of `AboutPage`; only visible in debug mode.

We came to the conclusion that `debug.js` can also serve as a temporary(!) place to put alternative implementations of js functions for testing.

Rebasing just got two conflicts. I am confident to have resolved them:
 - `ignoreEntities` in `getMessageText` and `enhanceMessageText`
 - `getSecretChatStatus`

I recommend merging this only after 0.5.1 nonetheless.

Wunderfitz:
Thanks for that (most certainly very tedious) work! Tests were successful, let's merge it! :)

PR #204, opened 2020-11-27
Fix initial auth handling

jgibbon:
This (hopefully, it at least LGTM) fixes two problems:
 - OverviewPage.onCompleted: It got the wrong AuthorizationState if qml was loaded too fast because it wasn't initialized. (c++)
- The QML could have tried to open the InitializationPage "onCompleted". But the initial pageStack operation is a parent in its call stack,  so it is neither ready nor wants to accept "completeAnimation()". A zero interval Timer was added for this case.

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -68,6 +68,14 @@ Page {
         }
     }
 
+    Timer {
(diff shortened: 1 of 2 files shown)


Wunderfitz:
OK, thanks!

Issue #156, opened 2020-11-13
Muted Chat icon inside text


jgibbon:
Currently, the muted text icon is integrated into text, so it may not be visible on low resolutions/long chat names. 
Also it changes horizontal alignment a bit.

Perhaps we could also stop interface-icon emojify-usage generally for performance reasons? 
We could
 - either just use the coloured icons used by emojify directly
 - or use system icons where applicable and create appoximations where none exist.

PR #205, opened 2020-11-27
New chat from contacts, initial secret chat support

### LINKED ISSUE #4: enable encrypted chats
Please add encrypted chat function including timer for self destroying messages! Hopefully Telegram API supports this!

### LINKED ISSUE #7: add "new chat" function
Basic function "new chat" is missing in V.0.1!

Wunderfitz:
OK, that's a big one and if it's merged it will fix #7 and partially fix #4 (no self-destructing messages yet). With this PR Fernschreiber will receive:

- A new menu option "New Chat" on the overview page
- A new chat page that lists all your known contacts in Telegram
- An option to synchronize local contacts with Telegram (see pulley menu on 'New Chat' page). In future we could think of doing the synchronization automatically if the user opts-in for that (I wouldn't do it automatically to protect the user's contacts)
- Tapping on a contact on the 'New Chat' page asks the user if a private or a secret chat should be created
- Secret chats should work in general, but I haven't tested it with many contacts, simply try it out ;)
- If a secret chat is closed, it will be added to the "hidden chats" and removed from the normal chat list

Thanks in advance for testing it. I'm particularly curious if everything works for you, what you think about the visuals that indicate if we have a secret chat or not and simply how you like the handling to initiate a new chat. Have fun!

P.S: If you wonder about the twemoji change: That fixed for me the display issues in texts with emojis. Theoretically it should be in a dedicated PR/commit, but I was just in the right mood to do it...


jgibbon:
Haven't had a thorough look at the sources yet, so I'll _mostly_ repeat my impressions from trying it a bit yesterday without a real review yet:
 - I tested it with an android device using my main account → sfos dev device (without contacts) using a test account. The secret chat worked really well. Good work!
 - If someone didn't want to sync their contacts (like me), an additional entry point to do a secret chat would be nice. As mentioned privately, I could imagine that either on the chat page (pulley?) or chat information page of a private chat.
 - I am a proponent of system iconography (or at least trying to emulate the style for custom ones), so I like the "lock icon" next to the profile thumbnail. The Emoji lock in the Chat title isn't quite my cup of tea: It's too big/colorful, it changes the (visual) horizontal alignment of the text a bit much and it probably isn't great on a some background colors. This is a bit related to #156.

cheers!

qml/components/PhotoTextsListItem.qml
```diff
@@ -10,6 +12,7 @@ ListItem {
     property alias tertiaryText: tertiaryText //usually last message date
 
     property int unreadCount
+    property bool isSecret: false
```
monich, line 15: `false` is a default for `bool`, just like zero for `int`

qml/components/PhotoTextsListItem.qml
```diff
@@ -50,6 +53,27 @@ ListItem {
                         height: parent.width
                     }
 
+                    Rectangle {
+                        id: chatSecretBackground
+                        color: Theme.overlayBackgroundColor
+                        width: Theme.fontSizeExtraLarge
+                        height: Theme.fontSizeExtraLarge
+                        anchors.left: parent.left
```
monich, line 61: I don't think `anchors.left: parent.left` makes any difference here, since default `x` is zero

qml/components/PhotoTextsListItem.qml
```diff
@@ -50,6 +53,27 @@ ListItem {
...
+                        anchors.bottom: parent.bottom
+                        radius: parent.width / 2
+                        visible: chatListViewItem.isSecret
+                    }
+
+                    Image {
+                        id: chatSecretImage
```
monich, line 68: Is this id used anywhere?

qml/components/PhotoTextsListItem.qml
```diff
@@ -50,6 +53,27 @@ ListItem {
...
+                        source: "image://theme/icon-s-secure"
+                        height: Theme.fontSizeMedium
+                        width: Theme.fontSizeMedium
+                        anchors.centerIn: chatSecretBackground
+                        visible: chatListViewItem.isSecret
+                    }
+
+
```
monich, line 76: Two empty lines?

qml/pages/ChatPage.qml
```diff
@@ -36,8 +36,11 @@ Page {
     property bool isInitialized: false;
     readonly property int myUserId: tdLibWrapper.getUserInformation().id;
     property var chatInformation;
+    property var secretChatDetails;
     property alias chatPicture: chatPictureThumbnail.photoData
     property bool isPrivateChat: false;
+    property bool isSecretChat: false;
+    property bool isSecretChatReady: false;
```
monich, line 43: Explicitly initialiazing properties to their default values looks weird. But consistent :)

qml/pages/ChatPage.qml
```diff
@@ -410,6 +428,22 @@ Page {
                 pinnedMessageItem.pinnedMessage = message;
             }
         }
+        onSecretChatReceived: {
+            if (secretChatId.toString() === chatInformation.type.secret_chat_id.toString()) {
```
monich, line 432: Are those `toString()` calls necessary?

qml/pages/ChatPage.qml
```diff
@@ -543,6 +577,19 @@ Page {
...
             visible: chatInformation.id !== chatPage.myUserId && !stickerPickerLoader.active && !messageOverlayLoader.active
+            MenuItem {
+                id: closeSecretChatMenuItem
+                visible: chatPage.isSecretChat && chatPage.secretChatDetails.state["@type"] !== "secretChatStateClosed"
+                onClicked: {
+                    var remorse = Remorse.popupAction(appWindow, qsTr("Closing chat"), (function(secretChatId) {
+                        return function() {
+                            tdLibWrapper.closeSecretChat(secretChatId);
+                        };
+                    }(chatPage.secretChatDetails.id)))
```
monich, line 588: Is there a reason for this being done in such a complicated manner? I'm not even sure how this is supposed to work. Remorse timers usually look much simpler than that.

qml/pages/ChatPage.qml
```diff
@@ -604,23 +651,49 @@ Page {
...
-                            chatPictureThumbnail.photoData = chatModel.smallPhoto;
+                    ProfileThumbnail {
+                        id: chatPictureThumbnail
+                        replacementStringHint: chatNameText.text
+                        width: parent.height
+                        height: parent.height
+
+                        // Setting it directly may cause an stale state for the thumbnail in case the chat page
+                        // was previously loaded with a picture and now it doesn't have one. Instead setting it
+                        // when the ChatModel indicates a change. This also avoids flickering when the page is loaded...
```
monich, line 668: Is it still the case or could `chatModel.smallPhoto` be bound directly to `chatPictureThumbnail.photoData` these days?

qml/pages/NewChatPage.qml
```diff
@@ -0,0 +1,394 @@
...
+                                width: parent.width
+                                height: contactListItem.height
+
+                                Item {
+                                    width: parent.width
+                                    height: parent.height - chatTypeSeparator.height
+
+                                    Rectangle {
+                                        anchors.fill: parent
+                                        opacity: 0.3
```
monich, line 194: Opacity `0.3` looks unusual. Pretty much all Sailfish OS UI elements have opacity `0`, `0.2`, `0.4`, `0.6`, `0.8` or `1`.

qml/pages/NewChatPage.qml
```diff
@@ -0,0 +1,394 @@
...
+                                        color: Theme.overlayBackgroundColor
+                                    }
+
+                                    Item {
+                                        id: privateChatItem
+                                        height: parent.height
+                                        width: parent.width / 2 // - ( Theme.horizontalPageMargin / 2 )
+                                        anchors.left: parent.left
```
monich, line 202: Unnecessary anchor?

qml/pages/NewChatPage.qml
```diff
@@ -0,0 +1,394 @@
...
+                                        anchors.top: parent.top
```
monich, line 203: This seems to be unnecessary too because of `height: parent.height`

qml/pages/NewChatPage.qml
```diff
@@ -0,0 +1,394 @@
...
+
+                                        Rectangle {
+                                            id: privateChatHighlightBackground
+                                            anchors.fill: parent
+                                            color: Theme.highlightBackgroundColor
+                                            opacity: 0.5
```
monich, line 209: Another unusual (inconsistent with SFOS UI style) opacity.

src/contactsmodel.cpp
```diff
@@ -0,0 +1,187 @@
...
+            LOG("Found contact" << singleContact.value("first_name").toString() << singleContact.value("last_name").toString() << singleContact.value("phone_number").toString());
+        }
+        if (!deviceContacts.isEmpty()) {
+            LOG("Importing found contacts" << deviceContacts.size());
+            this->tdLibWrapper->importContacts(deviceContacts);
+        }
+        emit contactsSynchronized();
+    } else {
+        LOG("Error selecting contacts from database!");
+        emit errorSynchronizingContacts();
```
monich, line 179: A strange way of returning a boolean value from a function by emitting a signal. Or did you mean that this function will later become asynchronous?

qml/pages/ChatPage.qml
jgibbon: Yes, if the page is closed, the variable is still available because it is "trapped" in the function scope.

monich: Wouldn't something like this do it?
```JavaScript
    var chatId = chatPage.secretChatDetails.id
    Remorse.popupAction(appWindow, qsTr("Closing chat"), function() {
        tdLibWrapper.closeSecretChat(chatId)
    })
```

src/contactsmodel.h
```diff
@@ -0,0 +1,61 @@
...
+#include "tdlibwrapper.h"
+
+class ContactsModel : public QAbstractListModel
+{
+    Q_OBJECT
+public:
+    ContactsModel(TDLibWrapper *tdLibWrapper, QObject *parent = nullptr);
+
+    virtual int rowCount(const QModelIndex &) const;
+    virtual QVariant data(const QModelIndex &index, int role) const;
```
monich, line 37: It's really really really really strange to see a list model that doesn't define any roles. Like, `firstName`, `lastName` and such. It's like a database that defines a single column called `data` and stores everything in that one column.

src/contactsmodel.cpp
```diff
@@ -0,0 +1,187 @@
...
+    while (userIdIterator.hasNext()) {
+        QString nextUserId = userIdIterator.next();
+        LOG("Hydrating contact:" << nextUserId);
+        this->contacts.append(this->tdLibWrapper->getUserInformation(nextUserId));
+    }
+    LOG("Hydrated contacts:" << this->contacts.size());
+    std::sort(this->contacts.begin(), this->contacts.end(), compareUsers);
+}
+
+void ContactsModel::applyFilter(const QString &filter)
```
monich, line 129: Consider using `QSortFilterProxyModel` for filtering and sorting. AFAICT it does a pretty good job in avoiding unnecessary updates of the view (let alone resetting the model on every change).

src/contactsmodel.h
monich: Which would of course work just fine - why would anyone need more than one column in a database, right? :) 

However, that would significantly limit options for all kinds of optimizations of updates, queries and other operations on that poor database. Same thing here. Yes, one role is enough, just like one column is enough for a database.

qml/pages/ChatPage.qml
Wunderfitz: Most certainly not... :)

qml/pages/ChatPage.qml
Wunderfitz: In this case I simply copied the logic over from the previous example by @jgibbon - if we change that, I'd like to do it for all of the similar examples...

qml/pages/ChatPage.qml
Wunderfitz: Don't know actually. With your recent change, the initialization is fine now, but I simply kept the onSmallPhotoChanged as I haven't had time to test all edge cases again...

qml/pages/NewChatPage.qml
Wunderfitz: Is that documented somewhere? But OK, will adjust it. :D

src/contactsmodel.cpp
Wunderfitz: It was originally intended to be asynchronous, but as it was fast enough, I kept it synchronous and forgot to adjust those thingies here. Will be changed...

src/contactsmodel.h
Wunderfitz: I was simply the easiest way of getting things done. In this simple thingy I don't really see the need to introduce additional roles, but of course it can be done... But wouldn't that be something that can be considered upfront overhead design? ;) Jokes aside, let me see if I find some time to implement roles here as well...

src/contactsmodel.cpp
Wunderfitz: Initially I wanted to implement it with `QSortFilterProxyModel`, but no matter what I did (applied a filter or not), I always ended up with an empty list. Maybe I'm simply too dumb to understand or once again used it in the wrong way and got stuck because of that. Anyway, I gave up and implemented it that way...

qml/pages/NewChatPage.qml
monich: Don't know about documentation but recent `Theme` defines constants `opacityFaint`, `opacityLow`, `opacityHigh` and `opacityOverlay` which are  `0.2`, `0.4`, `0.6` and `0.8`, respectively. If you don't care about compatibility with earlier versions of Sailfish OS (which you don't) then you can use those constants too.

src/contactsmodel.cpp
monich: Trust me, it does work :) If matches only one role though - if you want it to match several roles, then you would need to define additional "combined" role (not displayed in the UI) which would concatenate all roles involved in matching into one (separated with `'\n'` or something). But that obviously contradicts the "one role fits all" and "why would anyone need more than one column in a database" approach.

qml/pages/ChatPage.qml
jgibbon: I think (don't quote me on that,yet) I initially had the simpler version, because that would make sense and would work in "pure JS" or with primitive types, but IIRC it didn't.

If I get some time tonight (started self isolation today, many things to do/plan) I'll check again. 
I propose to leave it here as it is and if I confirm it's unnecessary, I'll make a separate PR (perhaps an issue).

monich: Sure, as long as it doesn't get forgotten.

It's just quite typical that something is initially being done in an overly complicated manner for some more or less valid reason, then that reason completely goes away but the complexity continues to be dragged along, contaminating other things along the way.

Life is a (hopeless) fight against entropy :)

qml/components/PhotoTextsListItem.qml
Wunderfitz: I actually like to get this type of overview - as you mentioned below, I try to make it consistent that way. ;)

qml/pages/ChatPage.qml
Wunderfitz: :)

qml/pages/NewChatPage.qml
Wunderfitz: I wasn't aware of these constants, thanks! I changed things accordingly and will change the other opacity values as soon as I touch the respective code areas...

src/contactsmodel.h
Wunderfitz: Yep, found some time. :)

src/contactsmodel.cpp
Wunderfitz: I went for a ride with the bike and for whatever reason it worked straight afterwards - though I'm close to 100% sure that I didn't do anything different than when I tried it last a few days ago. Anyway, we're using the `QSortFilterProxyModel` and I removed the duplicate filtered list...

Wunderfitz:
OK folks, I tried my best to accommodate most of the change requests (notable exception: variable initialization ;)). I even made the `QSortFilterProxyModel` work - which was really easy and I have no idea why it didn't work last time I tried it. Also found additional missing code for secret chats in the chat information page - which I added as well.

Have fun retesting!

qml/pages/ChatPage.qml
jgibbon: I have tested this now. 
And while it really didn't work before, the error is something different: 
For some (aka. apparently no) reason I called it with `appWindow` as a parent, so it just dies regardless of how complicated the rest is, so even that workaround is useless.
When using chatPage as it is intended, the simple solution works. Thanks for noticing!

I've created #206 to get rid of this.

PR #207, opened 2020-11-29
Animate opening and closing of attachment row

monich:
And slightly tweaked the process of closing the sticker picker.


Wunderfitz:
Thanks, that's nice! :)

PR #213, opened 2020-11-30
Update harbour-fernschreiber-ru.ts

arustg:
Update harbour-fernschreiber-ru.ts


Wunderfitz:
Спасибо!

However, one thing for the next time: Please avoid the standalone 'type' attributes - they cause parsing error. Don't worry, I'll fix it for now. :)

arustg:
👍 ok!

PR #219, opened 2020-12-01
Update functions.js

GNUuser:
#203

### DIFF
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -164,13 +164,13 @@ function getMessageText(message, simple, myself, ignoreEntities) {
 function getChatPartnerStatusText(statusType, was_online) {
     switch(statusType) {
     case "userStatusEmpty":
-        return qsTr("was never online");
+        return qsTr("never online");
     case "userStatusLastMonth":
-        return qsTr("offline, last online: last month");
+        return qsTr("last month");
     case "userStatusLastWeek":
-        return qsTr("offline, last online: last week");
+        return qsTr("last week");
     case "userStatusOffline":
-        return qsTr("offline, last online: %1").arg(getDateTimeElapsed(was_online));
+        return (%1) (getDateTimeElapsed(was_online));
     case "userStatusOnline":
         return qsTr("online");
     case "userStatusRecently":
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -164,7 +164,7 @@ function getMessageText(message, simple, myself, ignoreEntities) {
 function getChatPartnerStatusText(statusType, was_online) {
     switch(statusType) {
     case "userStatusEmpty":
-        return qsTr("never online");
+        return qsTr("was never online");
     case "userStatusLastMonth":
         return qsTr("last month");
     case "userStatusLastWeek":
@@ -174,7 +174,7 @@ function getChatPartnerStatusText(statusType, was_online) {
     case "userStatusOnline":
         return qsTr("online");
     case "userStatusRecently":
-        return qsTr("offline, was recently online");
+        return qsTr("was recently online");
     }
 }
 
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -170,7 +170,7 @@ function getChatPartnerStatusText(statusType, was_online) {
     case "userStatusLastWeek":
         return qsTr("last week");
     case "userStatusOffline":
-        return (%1) (getDateTimeElapsed(was_online));
+        return " %1" + (getDateTimeElapsed(was_online));
     case "userStatusOnline":
         return qsTr("online");
     case "userStatusRecently":


Wunderfitz:
What's the motivation behind this change?

Regards, Sebastian

GNUuser:
Those string I deleted is not useful, and mess up the Fernschreiber interface.

GNUuser:
I was tracking this issue from Fernschreiber v0.3 and already quick fixed in #218

Wunderfitz:
I'd disagree here - even the official client says something like "last seen recently". With this case, Fernschreiber would only say for example "two minutes ago" without anything else. I don't see why the current text would be too messy...

GNUuser:
Trust me  something is wrong there before my changes. Ln: 169, 171, 173 are not correct.

GNUuser:
In line: 173 may be is correct "last seen recently" but not "offline, last online:"

Wunderfitz:
There is a dedicated status for "last seen recently" and it's handled in https://github.com/Wunderfitz/harbour-fernschreiber/blob/master/qml/js/functions.js#L176 - See also the statuses from https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1_user_status.html

GNUuser:
offline, was recently online

> There is a dedicated status for "last seen recently" and it's handled in https://github.com/Wunderfitz/harbour-fernschreiber/blob/master/qml/js/functions.js#L176 - See also the statuses from https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1_user_status.html

You talk about line 177 I do not touched the line, but also need fix by removing "offline, was" .

GNUuser:
Check my es.ts history I see how I focus too much in those line, to get correct and optimized sentences.

Wunderfitz:
Well yes, I was wondering a bit why you changed certain strings back and forth, but that's OK. :) My Spanish is not that good that I can assess that and as long as there are no other people asking for a different translation, that's OK with me. If you want to change the original strings, however, you probably need to convince others as well. I'm really open to a shorter version, but that should be consistent and also still convey the respective status message.

Maybe let's better discuss that in the Telegram group...

GNUuser:
Sounds good, thank you. :  )

GNUuser:
Just remember that Fernschreiber it is not a book, It is a FANCY app like Telegram client to Sailfish OS running on an electronic device. Be careful not to put too many messages in plain text. ; )

jgibbon:
I think these changes lead to losing context a bit.
But I understand your problem and would propose a compromise. 
For example just removing "offline, " everywhere, since that is sufficiently implied with the rest of the strings.

Wunderfitz:
Yep, removing 'offline, ' only would be something I would also approve!

GNUuser:
Well some lines I deleted  only "offline".  We need to use only essential phrases, the English language its phrases are small, but other languages the phrases are long or medium, this is a problem when translating a string of words in portrait mode. In Spanish language it is extremely explicit and most of the sentences are long, so to translate Fernschreiber I am using advanced Spanish, to reach the correct size of the word string. Right now my translation is nice in newer version 0.5.1.
Issues examples
Truly -> Verdaderamente

monich:
+1 for replacing "offline, last online:"  with just "last online:"

Wunderfitz:
As the majority here was voting to only remove the 'offline' text, but nothing more, I did that with commit 633f765c8028ec40a7aabca741572fa3f88628ec. Moreover, I also changed the translations accordingly so that there are no additional 'unfinished' entries there.

@GNUuser I'll close this PR because of that change. As said, the Spanish translation is up to you (I haven't changed your shorter translation) unless there are other people asking to get closer to the original wording. 

Thanks colleages and best regards, Sebastian

GNUuser:
Mmm I find where is the problem, one string coming from directly from Telegram api, already do "offline, last online:"

GNUuser:
> As the majority here was voting to only remove the 'offline' text, but nothing more, I did that with commit [633f765](https://github.com/Wunderfitz/harbour-fernschreiber/commit/633f765c8028ec40a7aabca741572fa3f88628ec). Moreover, I also changed the translations accordingly so that there are no additional 'unfinished' entries there.
> 
> @GNUuser I'll close this PR because of that change. As said, the Spanish translation is up to you (I haven't changed your shorter translation) unless there are other people asking to get closer to the original wording.
> 
> Thanks colleages and best regards, Sebastian

Yes close it. Thank you.

GNUuser:
I am 110% confident that the issue or confusing is coming from the sentence(Ago) coming directly from Telegram api, so "offline, last online:" - > "Ago". I found the bug today, but I am okay with your changes, I just reporting to Fernschreiber  team.
Also I already readed the translations TelegramApi recommendations. @Wunderfitz @jgibbon @monich

PR #221, opened 2020-12-03
Add verification badge to verified chats

monich:
The attribute is extracted from the **supergroup** structure:

**supergroup** id:int32 username:string date:int32 status:ChatMemberStatus member_count:int32 has_linked_chat:Bool has_location:Bool sign_messages:Bool is_slow_mode_enabled:Bool is_channel:Bool **is_verified:Bool** restriction_reason:string is_scam:Bool = Supergroup;


monich:
Screenshot_20201203_001

jgibbon:
I like it! (Especially that it isn't inside the text, so it doesn't get elided with long stings.)

jgibbon (approved):

monich:
I was initially thinking of making a darker icon for light ambiences but decided that this one looks reasonably well as is:

Screenshot_20201203_004

Wunderfitz:
Yes, great. Works smoothly. Before we merge it: Please confirm that you own the copyright/you created the verified icon. I simply want to avoid issues because of such tiny things... ;)

Thanks and best regards, Sebastian

monich:
> Yes, great. Works smoothly. Before we merge it: Please confirm that you own the copyright/you created the verified icon. I simply want to avoid issues because of such tiny things... ;)

Yes, I sketched it myself in Inkscape, combining a check mark with two blue squares.

I wouldn't worry too much about that. It's either a generic shape which no one owns, or it's owned by someone in which case the fact that you have drawn the exact same (or very similar) shape yourself won't save you from being punished :)

Wunderfitz:
> Just because you're paranoid doesn't mean they aren't after you.

;) Merging...

PR #222, opened 2020-12-03
Improve preview of wide images

monich:
Avoid clipping. Also, handling file updates on C++ side should be good for performance.


monich:
This doesn't look right:

Screenshot_20201203_003

monich:
And this does:

Screenshot_20201203_002

Wunderfitz:
Very good idea, works smoothly! Thanks! :)

PR #224, opened 2020-12-04
Fixed long tap on an image

monich:
`MouseArea` filling the image was eating mouse events which should be handled by the list item. Handle them all at the list item level and forward the `"clicked"` event to the extra content items which declare the `clicked()` function.

This fixes context menu and background highlighting when image is long-tapped.


Wunderfitz:
Elegant solution, thanks!

PR #225, opened 2020-12-04
extend #224 to LocationPreview and VideoPreview

jgibbon:
#224 was a brilliant idea, I just did mostly the same for other components.


Wunderfitz:
Thx! :)

PR #226, opened 2020-12-04
Press effect for image preview items

monich:
Instead of this:

old-press

It should now be highlighted like this:

new-press


Wunderfitz:
Good addition, thanks! Probably it makes sense to apply that as well to other previews, but that's certainly not mission-critical.

monich:
> Probably it makes sense to apply that as well to other previews, but that's certainly not mission-critical.

Yes, I had that in mind, that's why `PressEffect` is a separate component.

PR #227, opened 2020-12-04
press effect for more items

jgibbon:
there are still a few areas where there's no press effect (message items themselves: Text, profile thumbnail, sent icon,…)
but for now I've just aligned the components to the image changes


Wunderfitz:
Thank you very much!

PR #229, opened 2020-12-05
Tweaked poll/quiz layout

monich:
Made it slightly more compact by aligning the percent label with the option text if there's enough space for it (which is the most typical use case, especially in landscape where vertical space is precious).

Instead of this:

poll-old

it looks like this:

poll-new

This also moves the option text closer to the associated percentage bar (rather than the one above it).


monich:
An example of the percentage label being pushed down if there's not enough space for it on the last line of the option text:

push-down

Wunderfitz:
Looks really nice, thanks for the improvements! :)

PR #232, opened 2020-12-06
Optimize chat model a bit

monich:
1. Store and handle message ids as numbers rather than variants/strings
2. Incrementally update message id map
3. Expose additional roles and properties to avoid unnecessary lookups


Wunderfitz (changes requested):
Thanks for the clean-up and the changes. However, it seems that at least deleting messages is no longer working ("Invalid message identifier"). Could you please check and change that accordingly?

monich:
Fixed and rebased

Wunderfitz (approved):
Looks good now. Thanks for the adjustment!

PR #234, opened 2020-12-06
Animate display of selection panel

monich:


Wunderfitz:
I love animations, thanks! :)

PR #236, opened 2020-12-15
Don't show Reply menu item if the message can't be replied to

monich:
Also moved the handling of the Edit action to the chat page where it's actually being handled.


Wunderfitz:
Ah, good catch! Thank you very much...

PR #237, opened 2020-12-15
Migration to TDLib 1.7

Wunderfitz:
As you might have heard - TDLib received [a major update to v1.7](https://github.com/tdlib/td/blob/master/CHANGELOG.md). 
With that update came some incompatible changes that Fernschreiber needed to be adjusted to. To be concrete:
- Message sender IDs (due to the support for anonymous authors in chats)
- Formatting in web page previews
- More consistent handling of files in objects (e.g. stickers, videos)
- Animated previews of animated GIFs (what's the sense in previews if they are already animated, but well...)

As I'd like to be up-to-date, some people already asked for the full support of anonymous messages in chats (might be done later) and in order to avoid huge efforts later when we really need to update to the latest version (e.g. if earlier TDLib versions no longer speak properly to the Telegram servers), I'd like to perform the migration better earlier than later. 

Thanks to the amazing folks at Telegram, especially @levlam, it seems that we also [avoid some strange Sailfish OS-related performance issues](https://github.com/tdlib/td/issues/1322) that I faced on my devices. Please find the complete `tdlib` folder with the ARM and i486 builds attached to this PR in case you don't want to build it on your own: [tdlib.tar.gz](https://github.com/Wunderfitz/harbour-fernschreiber/files/5698183/tdlib.tar.gz)

This PR only addresses the absolute necessary to get the same features we already had with v1.6 and to avoid migration bugs. Additional features that came with v1.7 may be implemented later. Have fun!

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ const char TDLIB_API_HASH[] = "1234567890abcdef1234567890abcdef";
 
 You get the Telegram API ID and hash as soon as you've registered your own application on [https://my.telegram.org](https://my.telegram.org).
 
-Moreover, you need to have a compiled version of [TDLib](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
+Moreover, you need to have a compiled version of [TDLib 1.7](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
+
 
 ## Debug
 Fernschreiber does only output a few TDLib messages by default. To get its own debug log messages, you can either run a debug build to see all of them or use the environment variable `QT_LOGGING_RULES` to specify/filter which messages you'd like to see.
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ You get the Telegram API ID and hash as soon as you've registered your own appli
 
 Moreover, you need to have a compiled version of [TDLib 1.7](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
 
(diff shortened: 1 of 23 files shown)


jgibbon:
As mentioned in private, I was only able to find one bug while quickly comparing the changes/tdlib change log:
On the ChatInfoPage, the big versions of the thumbnails aren't displayed anymore.

Obviously, there are lot of new features of which some may benefit from further changes, but obviously that is out of scope for this bare adaptation. The TDLib change log may be a good source for new feature request issues :)

src/tdlibreceiver.cpp
```diff
@@ -146,7 +146,7 @@ void TDLibReceiver::setActive(bool active)
 void TDLibReceiver::receiverLoop()
 {
     LOG("Starting receiver loop");
-    const double WAIT_TIMEOUT = 5.0;
+    const double WAIT_TIMEOUT = 1.0;
```
jgibbon, line 149: Was this a necessary change or just something you tried while debugging the delay and left in?
Just thinking about how this may or may not affect high latency connections…

Wunderfitz: Definitely a leftover, reverted it with the latest commit...

Wunderfitz:
Profile photos should work again, thanks for reporting!

README.md
```diff
@@ -42,7 +42,9 @@ const char TDLIB_API_HASH[] = "1234567890abcdef1234567890abcdef";
...
+Moreover, you need to have a compiled version of [TDLib 1.7](https://github.com/tdlib/td) in the sub-directory `tdlib`. This sub-directory must contain another sub-directory that fits to the target device architecture (e.g. armv7hl, i486). Within this directory, there needs to be a folder called `lib` that contains at least `libtdjson.so`. For armv7hl the relative path would consequently be `tdlib/armv7hl/lib`.
+
+In case you encounter strange performance issues on startup (several seconds delay, app seems to do nothing), please be sure to [follow the instructions from the respective GitHub issue](https://github.com/tdlib/td/issues/1322), i.e. let TDLib build SQLite with `-DOMIT_MEMLOCK` and be sure to comment the two lines 22558 (`#ifndef OMIT_MEMLOCK`) and 22567 (`#endif`) in the file `sqlite/sqlite/sqlite3.c`.
```
monich, line 47: Have you tried linking with `libsqlite3.so.0`? It's [allowed](https://github.com/sailfishos/sdk-harbour-rpmvalidator/blob/harbour-qa/allowed_libraries.conf) in Harbour.

Wunderfitz: No not really - I actually prefer to stay as close to the original library as it is. That makes it probably easier to get support. With that solution, the new version has worked smoothly since about a week now.

Wunderfitz:
I merge it now - my local tests were successful and also the previous issues with missing updates in channels are completely gone now. Moreover, no objections from anybody else. ;) In order to be able to continue on a common foundation (it's tedious to move between library versions when reviewing changes) I think it's better to move on now.

README.md
monich: Then you better reduce the default cache size like it's [done](https://git.sailfishos.org/mer-core/sqlite/blob/master/rpm/sqlite.spec#L83-84) in Sailfish OS. I'm quite sure that it's configured to be a memory hog by default.

PR #245, opened 2020-12-21
Removed unnecessary ampersand replacement

monich:
I don't think it would ever match anything.


Wunderfitz:
No matter what - I also removed this workaround in Piepmatz just yesterday: https://github.com/Wunderfitz/harbour-piepmatz/commit/ae1e9618deb35aa2a073a67b5ade309754ed6edc 
So agreed, let's get rid of it!

PR #247, opened 2020-12-24
Fixed Emoji image positioning in multiline StyledText

monich:
The problem - combination of `maximumLineCount` and `TruncationMode.Elide` (or `Fade`) breaks Emoji image alignment, pushing the image down, like this:

Explicitly truncating the text fixes the problem, at expense of certain runtime overhead. The image stays where it's supposed to be:

Also, added the code to toggle full and truncated Web page preview on tap. For example, tapping the above two things expands them into full preview:

Cool?


Wunderfitz:
Really cool, thanks! Honestly a bit surprised that the performance penalty is not that huge (didn't notice anything on my device), so let's merge it.

Issue #198, opened 2020-11-23
Prepare another optimization round


jgibbon:
I initially only thought about optimizing `OverviewPage`, because that didn't get much love, yet. But then some `ChatPage` issues "popped up", as well.

As I am not able to assess all of those points (not very familiar with how/why some of these things were done exactly, especially c++ interaction), I'd appreciate some feedback (or even implementation help later on) at least on the less obvious ones marked with :question::question:. They are numbered to make referencing a tiny bit easier.

## 1. OverviewPage
 - 1.1 Initialization is quite heavy: The `opacity` fade when loading gets false does stutter a bit. It will already be better if the other issues here are done. :question::question: Alternatively: Handle `loaded` in a `Timer` (maybe an existing one), as well?
 - 1.2 `updateContent()`
   - not a huge performance problem, but it gets called `onStatusChanged` which means also on `PageStack.pop`. :question::question: Do we need that for all calls inside? Or would it be enough to call (some of them?) on init?
 - 1.3 :question::question: Is `redrawModel` really, really needed? It recreates all rendered delegates AFAICT, which is a bit heavy & weird on init. Almost all other init issues' impact gets doubled by this one, I think.
 - 1.4 :question::question: `onChatChanged`: Doesn't do much, but it gets called a lot. Thousands of times on init for me. Do we really still need that?
   - :question::question: Also, does `chatListModel` even need to emit it multiple times on init?
 - 1.5 We should do a variant of the `ContextMenu` `Loader` workaround from `ChatPage`, as well. Profiler shows quite a bit of time wasted here.
 - 1.6 We should do a variant of `precalculatedValues` to reduce time spend on repeated (mostly `width`) bindings/calculations uselessly, which according to the Profiler adds up on init/while scrolling.

## 2. ChatPage
 - 2.1 viewMessageTimer: Only do `viewMessage` if the last read index < `lastQueuedIndex`?

### 2.2 MessageListViewItem
 - 2.2.1 `chatModel` `Connections` 
   - 2.2.1.1 `onMessagesIncrementalUpdate`: `chatModel.getLastReadMessageIndex` gets called for every message instance after every scroll
     - put `lastReadMessageIndex` into `precalculatedValues`, bind everything to that (also for the other signals)
     - :question::question: does `chatModel` need to emit it every time, even if nothing changed?
[...]

PR #248, opened 2020-12-25
speed up overview page loading

jgibbon:
contributes to #198

A few small changes to get the "time to chat list" pretty low:
 - put `ContextMenu` of `ChatListViewItem` into a `Loader` – the impact here isn't as heavy as on the `ChatPage`, but it saves a few ms on init and while scrolling.
 - radically reduced `interval` of `chatListCreatedTimer`. This just renders the list much, much sooner and seems to work pretty well.
 - put `tdLibWrapper.getRecentStickers()`,  `tdLibWrapper.getInstalledStickerSets()` and `tdLibWrapper.getContacts()` into a new Timer that is started when `chatListCreatedTimer` triggers. I don't think these are needed right at the start, so we can defer the needed time a bit.
 - removed `chatListModel.redrawModel()` when the chatList is created. In my tests, this workaround didn't seem needed anymore.
 - removed the `onChatChanged` `Connection` in every delegate.  In my tests, this workaround didn't seem needed anymore and even if it didn't do much, it got called a few thousand times on init in my setup, which added up. Also did this in `ChatInformationTabItemMembersGroups.qml`.
 - removed the main `Column`s. We basically only have a `PageHeader` (now containing the `GlassItem`) and the `chatListView` as layout items, so there's really no need for extra `Column`s calculating their own layouts. The loading indicator is now just anchored to the `chatListView`. 
 - Small layout change: I omitted the anchor `topMargin: Theme.paddingMedium` that would be needed on the `chatListView` to get the same layout as before. PageHeader has its own padding, so now the text seems better centered vertically.
 - replaced the `NumberAnimation`s with `FadeAnimation`s mainly because @monich likes that and I'm here to please.

~~One drawback~~ Two drawbacks I've noticed: 
- Sometimes I'm not able to read the loading text because it disappears too quickly. But I could get used to that.
- Updates/changes in the chat order may now be visible after initial load. I guess this is what the long interval was meant to hide, but I actually like it better this way – if you disagree, I'm happy to increase it again (maybe a bit lower than 300ms?).

As discussed quite a while ago, we'll have to test for a while if the two removed workarounds really aren't needed anymore or if I just got (un)lucky in the bug hunting lottery.

### DIFF
--- a/qml/components/ChatListViewItem.qml
+++ b/qml/components/ChatListViewItem.qml
@@ -24,39 +24,57 @@ PhotoTextsListItem {
     isSecret: ( chat_type === TelegramAPI.ChatTypeSecret )
 
     openMenuOnPressAndHold: true//chat_id != overviewPage.ownUserId
-    menu: ContextMenu {
-        MenuItem {
-            visible: unread_count > 0
-            onClicked: {
-                tdLibWrapper.viewMessage(chat_id, display.last_message.id, true);
-            }
-            text: qsTr("Mark all messages as read")
-        }
 
-        MenuItem {
-            visible: chat_id != listItem.ownUserId
-            onClicked: {
-                var newNotificationSettings = display.notification_settings;
-                if (newNotificationSettings.mute_for > 0) {
-                    newNotificationSettings.mute_for = 0;
-                } else {
-                    newNotificationSettings.mute_for = 6666666;
-                }
-                newNotificationSettings.use_default_mute_for = false;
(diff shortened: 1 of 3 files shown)


Wunderfitz:
Thanks for the work! Seems to work, but as you also said - some of the things will need to be tested in the long run. ;)

You are right about the intention of the `chatListCreatedTimer` - I suppose 20ms are not enough, but you're also right that 300ms are most certainly too much. Maybe I'll set it to 100ms later on...

PR #249, opened 2020-12-26
support basic bot messages (reply markup)

jgibbon:
Summary:
I've basically interacted with the official music bot ([at]music), implemented the easiest things that came to mind and fixed some errors on the way.

Change details:
 - `functions.js:` added `textEntityTypeBotCommand` to `Functions.enhanceMessageText`: This just returns the command extracted from the link. The return value is sent as a new message from `MessageListViewItem.qml` if a link is clicked on a message.
 - `ChatPage.qml`: removed a warning when `chatView.contentComponentNames[model.content_type]` wasn't available just because these things really bug me sometimes. Not really relevant to the PR topic.
 - `AudioPreview.qml`: Added a Label for performer/title and fixed a few warnings. AudioPreview still is far from being pretty, but I didn't change any layout things there.
 - `MessageListViewItem.qml`: Added sending of `textEntityTypeBotCommand` and a new Loader for `ReplyMarkupButtons.qml`.
 - `ReplyMarkupButtons.qml` (new): Display custom (to get them layed out in rows/columns as returned in the reply markup) Buttons. Only `inlineKeyboardButtonTypeCallback` and `inlineKeyboardButtonTypeUrl` are implemented as of now. Buttons for other types are greyed out.
 - cpp: implement `updateMessageEdited` signal from TDlib; update chat model if it changes.

Screenshot_20201227_002

cheers

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -55,6 +55,7 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/components/PinnedMessageItem.qml \
     qml/components/PollPreview.qml \
     qml/components/PressEffect.qml \
+    qml/components/ReplyMarkupButtons.qml \
     qml/components/StickerPicker.qml \
     qml/components/PhotoTextsListItem.qml \
     qml/components/WebPagePreview.qml \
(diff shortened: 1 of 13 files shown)


jgibbon:
(afaik this is roughly what depecher supported as well in terms of reply markup)

Wunderfitz:
Never interacted that much with bots, but as far as I can tell, it's working really well. Impressive! Also thanks for the fixes along the way! :+1:

Issue #143, opened 2020-11-11
Remove send new message button when "Send message by enter" is chosen


iamnomeutente:
One could maybe remove the send new message button icon when "Send message by enter" is selected and move the attachment icon right.

Wunderfitz:
Implemented by @santhoshmanikandan - closing this one. :)

Wunderfitz:
Good point by @jgibbon in #252, never thought about it as I don't use this option. So reopening this one. What about showing the bubbles if an attachment is added? Would probably only be a small change...

PR #252, opened 2020-12-30
Implements Remove send new message button when "Send message by enter" is chosen

santhoshmanikandan:
[Implements 143](https://github.com/Wunderfitz/harbour-fernschreiber/issues/143). Send Message Icon button will now be hidden by default if "Send message by Enter" option is activated.


Wunderfitz:
Thanks for that! :+1:

jgibbon:
I think this may need some more thought on how to get a good user experience. 
For example just selecting a file requires to tap the input field and pressing enter...

PR #254, opened 2020-12-30
Message API usage cleanup

monich:
1. Pass `chat_id` where appropriate
2. Pass `message_id` and `chat_id` (which are numbers) as numbers
3. Use pre-initialized QStrings more often
4. Don't pass numbers by const reference, it doesn't make sense
5. Removed some redundant const modifiers

For now usage of ids (strings vs numbers) is a bit inconsistent. I'm planning to gradually (and slowly) migrate the code towards passing and using numbers as numbers (as opposed to strings and variants) in an effort to somehow slow the rate of performance degradation.

### DIFF
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -171,7 +171,7 @@ ListItem {
     Connections {
         target: tdLibWrapper
         onReceivedMessage: {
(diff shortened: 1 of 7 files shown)


Wunderfitz:
Thank you very much! :+1:

PR #255, opened 2020-12-31
implement message drafts

jgibbon:
- Drafts are only saved (or removed) when the ChatPage is destroyed.
- When the draft is a reply, that is also saved.
- The chat order gets updated when a draft is changed.
- The draft message gets displayed on the OverviewPage and gets prefilled on the ChatPage.
- Note: The implementation does not actually create a real formatted text including entities, it just shoves the current text "in there".

I've had some issues with the ChatListModel not updating sometimes, but with doing the order update last, this seems to work.
Quite possibly I did some (c++) part horribly wrong and/or ineffective, so please do judge me – I'd love to understand more. ;)

cheers!

### DIFF
--- a/qml/components/ChatListViewItem.qml
+++ b/qml/components/ChatListViewItem.qml
@@ -11,15 +11,17 @@ PhotoTextsListItem {
         photoData: photo_small || ({})
     }
     property int ownUserId
+    property bool showDraft: !!draft_message_text && draft_message_date > last_message_date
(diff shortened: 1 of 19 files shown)


Wunderfitz:
Works really well, thanks! Some `if`s could probably be removed as they seem to be duplicates or catch something that doesn't need to be catched (`QVariant` usually features some nice default return values), but that's OK: :) Let's merge it. If we encounter any issues, things can be updated later.

PR #257, opened 2020-12-31
Apply opacity to picture and badge as a group

monich:
It matters when page is being dimmed. Before:

before

and after this change:

after

Better?


Wunderfitz:
Yep, better. Thanks! :)

PR #259, opened 2021-01-01
Implement TextArea focus after message send option

### LINKED ISSUE #149: Prevent keyboard disappearing after sending a message
Perhaps it would be better if the keyboard wouldn't disappear when a message is sent.
Especially if "send with enter" is selected it is quite uncomfortable like this in my opinion 
I think after clicking on the field to write a message it should never disappear until you click elsewhere on the screen independent from "send with enter" or not.

jgibbon:
This was low hanging fruit, should be pretty self-explanatory and fixes #149

Happy new year!


Wunderfitz:
Thank you very much and a happy new year to you as well! :+1:

PR #263, opened 2021-01-02
Remove magnifying glass, reorganize title bar, introduce interaction hint

Wunderfitz:


jgibbon:
The interaction hint is a good idea! 
Also, thanks for investing your time mostly just because I disliked the icon position. 
Very much appreciated!

A few remarks, though, to possibly make it _even better_:
- I think I agree with @monich – `anchors.fill: parent; invert: true` on the hint doesn't look as cut off and retains a bit more visual connection to the header.
- All elements (`pageHeader`, `chatSearchField`, `titleInteractionHint`) are switched directly by setting `visible` to `true`/`false` – if they'd get switched via `opacity` and get a `visible: opacity > 0`, the opacity animations would play in both directions which looks a bit smoother.
- The row became obsolete now, it can just be removed when `chatListView` is anchored bottom to `pageHeader`.

You may try it with these points adressed on 728862873cfc63f5eda94fd8b99441b151972639 from https://github.com/jgibbon/harbour-fernschreiber/tree/lean-overview (and if you agree, feel free to just pull to this branch ;) )

cheers!

PR #266, opened 2021-01-02
Add a different placeholder text for no filtered chats

jgibbon:
Another really small one.

Also adds a description to look in the pulley for new chats.


Wunderfitz:
Ah, good one, thanks!

PR #267, opened 2021-01-02
'Fernschreiber' is not localizable

GNUuser:
Fixed conflict with name.desktop
I really do not know what to do with apps with generic name.  For me these applications with generic names need to be translated as core-applications, but I don't know what the original author's wishes are.

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -124,7 +124,7 @@ Page {
             break;
         case TelegramAPI.ConnectionReady:
             pageStatus.color = "green";
-            pageHeader.title = qsTr("Fernschreiber");
+            pageHeader.title = Fernschreiber;
             break;
         case TelegramAPI.Updating:
             pageStatus.color = "lightblue";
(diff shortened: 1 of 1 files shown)


Wunderfitz (changes requested):
I wouldn't call it a 'conflict'. At the time of building that, I wasn't really sure if the name shall be localized or not - probably it won't so it makes sense to remove the `qsTr` here. However, if you really want to clean things up, please do it consistently. You probably also need to change https://github.com/Wunderfitz/harbour-fernschreiber/blob/master/qml/pages/OverviewPage.qml#L127 and the localization files.

Wunderfitz (changes requested):
Please put the name "Fernschreiber" in quotation marks, it's a string and not a variable. Moreover - if possible - please also adjust the localization files. You can let the Sailfish IDE do that automatically by starting a new build with the changed source file.

GNUuser:
I canceled this PR, then you fix to me. I don't know much about the code of Ferni.

PR #268, opened 2021-01-03
Support sending voice notes and locations

### LINKED ISSUE #238: enabling "send a voice note"
It would be great to be able to send voice notes!

Wunderfitz:
This change adds two more options to the attachment row when authoring new messages - voice notes and locations, so it fixes #238. There is probably still a lot of potential for improvement, but it seems to work quite well for an initial support. ;)

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -16,7 +16,7 @@ CONFIG += sailfishapp sailfishapp_i18n
 
 PKGCONFIG += nemonotifications-qt5 zlib
 
-QT += core dbus sql multimedia
+QT += core dbus sql multimedia positioning
 
 DEFINES += QT_STATICPLUGIN
 
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -58,6 +58,7 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/components/ReplyMarkupButtons.qml \
     qml/components/StickerPicker.qml \
     qml/components/PhotoTextsListItem.qml \
+    qml/components/VoiceNoteOverlay.qml \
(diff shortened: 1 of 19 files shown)


jgibbon:
Looks very nice to me, good work. This is going to make some users quite happy!

I don't really have reliable GPS on sfos, so I haven't had a comprehensive look at that, but sending the "rough" location the system got from wherever worked. Voice notes worked, as well.

I think this is fit for merging as is, but I still have a few suggestions (nitpicking, really):
 - while recording, the length indicator starts with a bit of delay if I saw correctly, also when re-recording, it stays at the length of the previous recording for a second. Resetting to 00:00 directly would be a nice touch.
 - while re-recording a voice note, the "use recording" button is a bit confusing – I think it could just be hidden while recording.
 - the attachment row content on my device (Xperia X) has "exactly the wrong width" (weird summary, I know): It looks almost nicely centered, so it doesn't look like it's scrollable at first glance, but the image button is out of view. 
 - scrolling the attachment row clips the icons at horizontalPageMargin, which does not look great.

I've tried to address the last two points by just using a `OpacityRamp` "downstream". It's not perfect as it fades one side at most, but I think it does the job well enough without being _too_ expensive:
 - https://github.com/jgibbon/harbour-fernschreiber/commit/5467897ce7adbe44526ee59c9106413fbc3c6547
 - https://raw.githubusercontent.com/jgibbon/harbour-fernschreiber/5467897ce7adbe44526ee59c9106413fbc3c6547/qml/pages/ChatPage.qml

The fade values roughly correspond to those that would be used by a faded `Label` of the same width.
As usual: Feel free to use or discard as you wish. ;)
 
cheers!

Wunderfitz:
Thanks for the testing and the suggestions. I'll integrate them this evening...

PR #283, opened 2021-01-05
Update harbour-fernschreiber-zh_CN.ts

dashinfantry:
improve cover layout


Wunderfitz:
Let's see how it goes...

PR #284, opened 2021-01-05
Update harbour-fernschreiber-zh_CN.ts

dashinfantry:
improve a string


Wunderfitz (changes requested):
It seems that you used an old file as foundation which deletes quite some strings, set some brackets the wrong way and also removed some singular/plural definitions. Please revise!

Issue #291, opened 2021-01-06
Error : 429 : Too Many Requests: retry after 1


monich:
Sometimes when you open a chat, Fernschreiber goes into a loop of some sort consuming lots of CPU and printing tons of messages like this:
```
[ 2][t 4][1609955073.875465631][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.876463651][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.877481460][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.893348217][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.894533395][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.895529031][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.896537542][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.897451400][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[ 2][t 4][1609955073.908968448][FileManager.cpp:3585][#1][!FileManager]	Failed to DownloadWaitFileReference file 8183(0) of type photos: [Error : 429 : Too Many Requests: retry after 1]
[...]

PR #296, opened 2021-01-08
Don't retry downloading too often

monich:
At very least this should significantly reduce the system load when the problem https://github.com/Wunderfitz/harbour-fernschreiber/issues/291 occurs.

### DIFF
--- a/src/tdlibfile.cpp
+++ b/src/tdlibfile.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (C) 2020 Slava Monich at al.
+    Copyright (C) 2020-2021 Slava Monich at al.
 
     This file is part of Fernschreiber.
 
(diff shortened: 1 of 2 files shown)


jgibbon:
I haven't found a useable one (error just has a message string which could change without further notice), but perhaps there's a way to get the actual wait time?

monich:
> I haven't found a useable one (error just has a message string which could change without further notice), but perhaps there's a way to get the actual wait time?

AFAICT, no messages are coming in from tdlib (other than `"file"` and `"updateFile"`, there's a sample trace [here](https://github.com/Wunderfitz/harbour-fernschreiber/issues/291#issuecomment-756457434)), at least not through the JSON interface.

Wunderfitz:
Solution sounds reasonable, thanks for taking care!

PR #297, opened 2021-01-08
Update webPage photo id when message gets updated

monich:
This only takes care of  `webPage`, something similar needs to be done for other components which download files.

Additionally, this fixes the problem with `ChatModel` roles not being updated. One needs to pass the list of changed roles to  `dataChanged` signal, then it works nicely. `ChatModel::messageUpdated`  signal is probably not needed anymore.


Wunderfitz:
Thank you very much, seems to work smoothly. :)

PR #303, opened 2021-01-10
Apply press effect to profile thumbnail in chat list?

monich:
If this item is pressed:

highlight-5

it currently looks like this:

highlight-6

and after this change the profile image (or the replacement circle) looks pressed too:

highlight-7


Wunderfitz:
Yep, let's do it. Thanks! :)

PR #307, opened 2021-01-10
Implement inline queries (among others)

### LINKED ISSUE #91: Implement functionality of telegrams @gif-Bot
In the official Android- and Desktop-App there is the @gif functionality. Typing that, a subset of GIF-Images is opened. With typing text after @gif, the available GIF's match that keyword.
I would love to see that function in Fernschreiber.

jgibbon:
This PR fixes #91 

## summary of changes

This PR aimed to just add `InlineQuery` support to Fernschreiber. 
But in the process, a few more things turned up. I'm sorry, I really should've done some of those separately.

This is what is basically the change log for this PR:
 - [feature] Implements inline queries including all 12 `inlineQueryResult` types.
 - [feature] handle reply markup ("bot buttons") inside `MessageOverlayFlickable`.
 - [feature] handle via_bot_user_id in `MessageListViewItem` and `MessageOverlayFlickable` (prefills message text on click if it's an inline bot)
 - [feature] don't append "openDirectly" automatically in `tdLibWrapper.createPrivateChat` (and the other `create*Chat` methods and after `searchPublicChat`) to allow for other uses.
 - [feature] implement `messageGame`
 - [feature] implement `messageGameScore` (simple layout)
 - [feature] implement new reply markup buttons: 
   - `inlineKeyboardButtonTypeCallbackGame`
   - `inlineKeyboardButtonTypeSwitchInline`
   - `keyboardButtonTypeText`
 - [feature] show bot description as `chatView.header` in chats with a bot.
 - [feature] don't `handleMessageTextReplacement` on ChatPage if `TextArea` isn't focussed
 - [bugfix] Check for correct result inside `LocationPreview`
 - [bugfix] Changes attachmentPreviewRow content/visibility handling
   - `fileProperties` were checked with `typeof === "undefined"` but set to `{}`. Now (together with `locationData`) it will be set to `null`
   - visibility is now `(!!locationData || !!fileProperties) && !inlineQuery.active`
  
## help for testing

While it's easy to find inline bots to use for some types, for others that's not the case. 
For these I switched my existing testbot to a really, really simple inline bot implementation temporarily.

Just go to a random chat and type "@VeloxTestBot showbutton". Clicking on the appearing button should bring you to a private chat where it'll present you with some other buttons to click. You should be able to see the bot description on top of the chat messages.
 - The first two buttons test prefilling the TextArea in a different chat (once with extra text) to start an inline query.
 - The following ones just prefill in the current chat to test some inline content types.
 - Hint: If you press the "X" icon if there's the bot mention + query string, only the query string is removed on first press. You can long press to clear the TextArea completely.
 - The test bot does not have multiple pages of results – to test loading more results, go on to the next steps.

After clicking through those, there are only a few content types missing. 
For those, you can just type "@gif [something]" (`inlineQueryResultAnimation`), "@music " (`inlineQueryResultAudio`) "@gamebot " (`inlineQueryResultGame` – and the games themselves), "@sticker [some emoji]" (`inlineQueryResultSticker`) and "@pic [something]" (`inlineQueryResultPhoto`).
"@foursquare" should do something with shared location, but that is a bit wonky still – see notes.

## please note:
  - Animation/Video: Playing all Videos at once was a sure way to get all media playback on my device fracked until restart. For that reason, only thumbnails get displayed. A click will show a Dialog where the Video is played and can be selected.
  - Audio: Playing an audio file pauses a playing audio file from another delegate
  - Stickers: Stickers aren't animated by default for performance reasons. If it is an animated sticker, a small indicator icon is displayed. For these stickers you can preview the animation by long pressing. While pretty hidden, I really like this workaround. It's almost an easter egg.
  - Voicenotes: I couldn't get my bot to get voicenotes really working. In some cases, telegram decided to send it as a document, in others the files just plain didn't work. Looks OK, though.
  - Location: Implementation of sharing the location with inline bots is naive at best. As mentioned before, I don't really have reliable location services on SFOS. I didn't see that as high priority at all, though.

### DIFF
--- a/qml/components/GamePreview.qml
+++ b/qml/components/GamePreview.qml
@@ -1,0 +1,126 @@
+/*
+    Copyright (C) 2020 Sebastian J. Wolf and other contributors
+
+    This file is part of Fernschreiber.
+
+    Fernschreiber is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
+    Fernschreiber is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with Fernschreiber. If not, see <http://www.gnu.org/licenses/>.
+*/
+import QtQuick 2.6
+import Sailfish.Silica 1.0
+import WerkWolf.Fernschreiber 1.0
+import "../js/functions.js" as Functions
+import "../js/twemoji.js" as Emoji
+
+Column {
+    id: gamePreviewItem
+
+    property ListItem messageListItem
+    property MessageOverlayFlickable overlayFlickable
+    property var rawMessage: messageListItem ? messageListItem.myMessage : overlayFlickable.overlayMessage
+    property bool highlighted
+
+    width: parent.width
+    height: childrenRect.height
+
+
+    Label {
+        width: parent.width
+        font.bold: true
+        font.pixelSize: Theme.fontSizeSmall
+        text: Emoji.emojify(rawMessage.content.game.title || "", font.pixelSize)
+        truncationMode: TruncationMode.Fade
+        textFormat: Text.StyledText
+        wrapMode: Text.Wrap
+    }
+    Label {
+        width: parent.width
+        font.pixelSize: Theme.fontSizeExtraSmall
+        text: Emoji.emojify(rawMessage.content.game.description || "", font.pixelSize)
+        truncationMode: TruncationMode.Fade
+        textFormat: Text.StyledText
+        wrapMode: Text.Wrap
+    }
+    Label {
+        width: parent.width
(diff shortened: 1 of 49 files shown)


qml/components/GamePreview.qml
```diff
@@ -0,0 +1,126 @@
...
+        width: parent.width
+        height: Theme.paddingLarge
+    }
+
+    Image {
+        id: thumbnail
+        source: thumbnailFile.isDownloadingCompleted ? thumbnailFile.path : ""
+        fillMode: Image.PreserveAspectCrop
+        asynchronous: true
+        width: visible ? parent.width : 0
```
monich, line 78: This is strange. First, invisible items should be completely ignored by the Column, their size doesn't matter. Second, what sets `visible` for this item? Did you forget `visible: opacity > 0`?

jgibbon: You are absolutely right on both accounts. I mixed this up when changing my initial anchor based layout to a less sophisticated/easier column. 
Thank you!

Wunderfitz:
Really incredible work! Thank you very much, discovered so much stuff, I even didn't know it was existing... Seems to work well - and even if there are issues, they can be fixed later on. :)

fridlmue:
Wuhuuu, nice! Thanks a lot!!!

PR #311, opened 2021-01-12
Log out and log in as a different user

### LINKED ISSUE #71: Log out and log in as a different user
Need to be able to logout (deleting the history) and log in as a different user.

santhoshmanikandan:
Added option to logout from the current user account under About page with Remorse timer option. 

Ref: #71 

PS: Doesn't add Multi-Account support yet

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -162,6 +162,7 @@ Page {
             }
             break;
         case TelegramAPI.AuthorizationReady:
+            loadingBusyIndicator.text = qsTr("Loading chat list...");
             overviewPage.loading = false;
             overviewPage.initializationCompleted = true;
             overviewPage.updateContent();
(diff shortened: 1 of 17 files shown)


qml/pages/OverviewPage.qml
```diff
@@ -368,6 +384,21 @@ Page {
                 size: BusyIndicatorSize.Large
             }
         }
+
```
jgibbon, line 387: Perhaps you could use/change the existing loadingLabel (with a different text, obviously) to reduce the amount of qml items?

jgibbon:
Nice work! 
Didn't do a real built/test, just did a quick scroll through the diff, but from that, it seemed pretty good.

qml/pages/OverviewPage.qml
santhoshmanikandan: Thanks for the suggestion @jgibbon, I have merged the two Busy Indicators to one now.

Wunderfitz:
Seems to work really well, thank you very much!

PR #316, opened 2021-01-13
Handle message data change in ImagePreview

monich:
And use `TDLibFile` to load image in `ImagePage`.

That seems to fix a problem with some images never getting loaded until I reload the chat page but I'm not 100% sure about that 🙂


jgibbon:
Looks good to me and works well! 
Also not 100% about the not loading part, but didn't encounter it while testing.

I'd merge that.

Wunderfitz:
Yep, looks good. Thank you very much!

PR #318, opened 2021-01-13
Show channel members on chat information page

monich:
You're allowed to get the list if you're channel's creator/admin. In that case you get `"can_get_members": true` in `"@type": "supergroupFullInfo"` message.


jgibbon:
That's a effective, short addition and works really well.
I missed the "can_get_members" permission – thanks for noticing and fixing!

@Wunderfitz I think this can be merged without hesitation.

Wunderfitz:
Thank you very much @monich and @jgibbon - will do that! :)

PR #319, opened 2021-01-14
Display chat title in SimpleMessageDelegate if sender is chat

jgibbon:
On overviewPage an empty sender name for channels looks alright since the title is directly above it, anyway. 
But in the channel itself, the "simple messages" seemed wrong without a sender name.

The name link doesn't have a href in that case – this would only be useful for edge cases like someone sharing a "changed title" message to another chat with some inofficial client/bot that supports this. I did not consider that relevant enough. If other use cases arise, that can be added later.

image


Wunderfitz:
Good catch, thanks!

PR #323, opened 2021-01-15
Message content refactor

jgibbon:
This PR creates a dedicated file for every supported messageContent type and introduces a base type to handle properties that every Item must have, as well.
As proposed in the dev group, this simplifies QML lookups and (at least I think so) should ease getting started for new contributors a bit.

If I haven't missed something, I've tried all content types at least once after these changes.

- Items using the same QML as another one just inherit the other one (e.g. MessageAnimation.qml → MessageVideo.qml).
- I have changed "barely anything" inside the renamed Items to ease reviewing. Further planned changes (like now unneccessary conditionals for different types that can be solved by just presetting different values or, of course, using the previously discussed "global" items for image/photo types) will be prepared in dedicated PRs after one another for the respective items. 
- Exception: I hot-fixed some stuff for locations and changed the image width for games, but those will still get some "real love" in the near future, as well. Since the reuseable "TDLibImage" qml isn't included yet, those are still mostly the same old ugly things just to get it working again quickly.
- Because WebPreview is strongly related, I have included it in the newly created folder.
- For some reason, "git mv" didn't retain the history for MessageGame. It should've.


jgibbon:
You may disregard some height presets not matching up yet – that will be touched upon in the follow-up PRs.

jgibbon:
As discussed, merging this now to continue with optimizing or beautifying the individual content types

PR #325, opened 2021-01-17
Fix QML runtime warning in ChatInformationTabItemMembersGroups

monich:
When opening chat info page with member list:
```
[W] unknown:75 - file:///usr/share/harbour-fernschreiber/qml/components/chatInformationPage/ChatInformationTabItemMembersGroups.qml:75:28: Unable to assign QString to QVariantMap
```
This change makes that warning go away.


jgibbon:
Nice. I guess we may still have a few of those left-overs. 

I tried it, it works well.

jgibbon:
There's no reason not to merge this, so I think I might as well save @Wunderfitz that one click. ;)

Thanks!

PR #327, opened 2021-01-17
Add TDLibImage/TDLibThumbnail; rework Audio/VoiceNote/Document

### LINKED ISSUE #308: Document messages layout
Right now, document messages have a functional design but it's rather text/button heavy.

 - Perhaps there's a good way to use iconography and less text/fewer buttons. We'll have to differentiate between "download" (enable opening) and "copy" (to downloads folder).
 - Maybe just using a progress circle could be less obtrusive than the bar.

jgibbon:
also, again, a quick Location fix before its real turn. This fixes #308.

This PR has two main parts:
 - Creating a unified base for the `messageContent` types audio, voiceNote and document. The main goal for this is to have a small and consistent vertical footprint for these items. The audio types get bigger while playing but still only have a small height of `Theme.itemSizeLarge` like the document one when scrolling over them. I also wanted to make them a bit more visually pleasing.
 - Introducing and using TDLibImage and TDLibThumbnail (The Photo equivalent will probably be in the next PR ;) ). In my initial tests, these worked well and should be relatively easy to put… almost everywhere. It's going to be some work – we even still have a lot of places where even TDLibFile isn't used – but has the potential to reduce duplicated code in a lot of places.

Example: Downloaded document. Primary text gets a bit smaller if it's too long. For downloaded documents, the clickable area to open the file spans the file type icon and the labels. The folder icon is for copying the file to the local downloads folder.
1b4cbc17-dedd-42c7-be49-28dd6610368e
Example: Document. (Even if it is an audio file.) Click on the "File type icon with cloud" downloads the file.
65c23e5e-5fb2-4f05-8f0d-bedba9c5d703
Example: Downloaded audio. The bright square is the telegram provided thumbnail (dimmed with opacity).
86117874-ddff-4548-a342-6cc0a603b063


jgibbon:
Oh, totally forgot: The screenshots have the old folder icon in them. The new one is based upon an idea by @monich  ;)

Wunderfitz:
Really, really good new UX. Also functionality-wise I haven't found any issues, I'll simply merge it. In case we find anything we can still correct it.  :) Thank you very much!

PR #328, opened 2021-01-18
Visually optimize chat list

jgibbon:
- work on pin icon (mirrored/pointier)
- align profile thumbnail the same as the text fallback
- harmonize list delegate (and profile thumbnail) height (no more variable height)
- minimally reduce column spacing
- change pin/secret indicator background

The actual difference in layout (except for the indicators, of course) is practically invisible if you don't know it happened. 
Chat photos are a bit bigger now (Theme.itemSizeLarge instead of the variable size) and the list delegates are about 2 px taller.


jgibbon:
By popular demand, I changed the pin head to be more hat shaped.

Wunderfitz:
Looks really much more compact than before, but I like it. Thank you very much - of course also for the icon adjustment.

PR #329, opened 2021-01-18
Change channel messages layout

jgibbon:
- make text wider
- add view count

channel


Wunderfitz:
Nice one, thank you very much! :)

PR #330, opened 2021-01-19
Feature/message content photo

jgibbon:
This updates a few items for message content to use the new abstraction elements, mostly messagePhoto.
If available, a blurred ultra low res thumbnail is shown while loading many items. (Probably more to come ;) )

Additional changes:
 - images have a minimum height now to ensure clickability
 - precalculated height of those smaller images should match the real one now (less scroll flicker, yay!)
 - animations also have less scroll flicker
 - location/venue crept into this pr and uses TDLibImage now
 - image page shows the biggest photo size now


Wunderfitz:
Thank you very much, that looks really nice! :+1:

PR #332, opened 2021-01-22
Poll changes

jgibbon:
- fixes the poll results page
- implements quiz explanation (TDLib 1.7 feature)


Wunderfitz:
Thanks for the fix and extension!

PR #336, opened 2021-01-24
ChatListModel experiment

jgibbon:
**It may not be a good idea to merge this. It is mainly an "academic" testbed and an opportunity to both learn stuff and perhaps discuss this.** 
_If it turns out that these changes actually make sense, I'd be glad to be wrong, though. ;)_

There are a few things I learned and tried to adress. 
We:
 - basically have all known chats fully cached both in `ChatListModel` and `TDLibWrapper`. I don't know why. I've tried to use `TDLibWrapper::getChat()` for the `ChatPage`, for example, and that worked, as well. (Pinned message handling wasn't used outside that cache, as well, so it was moved to `TDLibWrapper`, also items in the `TDLibWrapper` cache get updated now.)
 - really don't need the `ChatData::chatData` in `ChatListModel` – we only need a few more dedicated variables instead of getter functions for the `QVariantMaps`.
 - may not even need the `hiddenChats` cache. Not sure, though, how those are actually used. It's still included in these changes.

You probably want to see some results. The memory footprint is a bit different but I wouldn't call it a night and day difference (I've mostly got ~1MB fewer allocations when comparing the maximum measurements).

Here's a somewhat average example of the five tests I did for before and after, both initializing and slowly scrolling the chat list down once:
image
image

Also I think these changes may result in a few less "Variant value conversion" function calls. On the first few glances, things still work, but… I may have broken some, nonetheless. _That is why there's that disclaimer on the top._ 

I know you guys most likely don't have much time to really review this in detail, but it is mostly what _I think_ @monich meant when he listed things to avoid with QVariantMap. 

So, what do you think? ;)

### DIFF
--- a/src/chatlistmodel.cpp
+++ b/src/chatlistmodel.cpp
@@ -79,7 +79,6 @@ public:
     TDLibWrapper *tdLibWrapper;
 
 public:
-    QVariantMap chatData;
     qlonglong chatId;
     qlonglong order;
     qlonglong groupId;
@@ -104,7 +103,6 @@ public:
 
 ChatListModel::ChatData::ChatData(TDLibWrapper *tdLibWrapper, const QVariantMap &data) :
     tdLibWrapper(tdLibWrapper),
-    chatData(data),
     chatId(data.value(ID).toLongLong()),
     order(data.value(ORDER).toLongLong()),
     groupId(0),
@@ -580,7 +578,7 @@ void ChatListModel::updateSecretChatVisibility(const QVariantMap secretChatDetai
         if (chat->chatType != TDLibWrapper::ChatTypeSecret) {
             continue;
         }
-        if (chat->chatData.value(TYPE).toMap().value(SECRET_CHAT_ID).toLongLong() != secretChatDetails.value(ID).toLongLong()) {
+        if (tdLibWrapper->getChat(chat->chatId).value(TYPE).toMap().value(SECRET_CHAT_ID).toLongLong() != secretChatDetails.value(ID).toLongLong()) {
             continue;
         }
         const QVector<int> changedRoles(chat->updateSecretChat(secretChatDetails));
--- a/src/chatlistmodel.cpp
+++ b/src/chatlistmodel.cpp
@@ -63,11 +63,6 @@ public:
 
     int compareTo(const ChatData *chat) const;
     bool setOrder(const QString &order);
-    const QVariant lastMessage(const QString &key) const;
-    QString title() const;
-    int unreadCount() const;
-    QVariant photoSmall() const;
-    qlonglong lastReadInboxMessageId() const;
     qlonglong senderUserId() const;
     qlonglong senderChatId() const;
     bool senderIsChat() const;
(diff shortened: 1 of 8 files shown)


monich:
I didn't even know there was a copy of chat model inTDLibWrapper :/ It definitely makes sense to get rid of it.

jgibbon:
It's not really a model inside TDLibWrapper, just a Map. But yeah.
I did go the other way here and left the TDLibWrapper one – possibly a bad decision? 
My thought was to (eventually) only use inside the model what's needed there, while the wrapper has multiple purposes used elsewhere.

On the other hand, we possibly wouldn't need a complete cache at all if we managed to use the tdlib methods to asynchronously get chat data everywhere on demand.

src/chatlistmodel.cpp
```diff
@@ -795,31 +761,15 @@ void ChatListModel::handleChatPhotoUpdated(qlonglong chatId, const QVariantMap &
...
+            chat->photoSmall = photo.value(SMALL);
         }
     }
 }
 
 void ChatListModel::handleMessageSendSucceeded(qlonglong messageId, qlonglong oldMessageId, const QVariantMap &message)
 {
     bool ok;
-    const qlonglong chatId(message.value(CHAT_ID).toLongLong(&ok));
+    qlonglong chatId(message.value(CHAT_ID).toLongLong(&ok));
```
monich, line 772: I don't like `this->`, somebody doesn't like `const` :)

I would like to point out though, that unlike `this->`, `const` is functional. It a) makes sure that you don't update a variable which is not supposed to be updated and b) gives the compiler a chance to better optimize the code.

jgibbon: Oh yeah, I stumbled a bit there. 
This was mainly to get rid of warnings where I used it in the following lines. If we end up actually using this, I'll be sure to to adress this. Thanks for the explanation!

monich: Not to mention that marking an instance of a Qt collection (QMap, QList and such) as const makes sure that non-const methods don't get accidentally invoked and shared content doesn't get unnecessarily copied (or "detached" as it's sometimes called).

monich:
It definitely makes sense to have it (the model, map, whatever you call the data structure describing the chats known to the client) in one place. And since you need the list model anyway, ChatListModel seems to be the natural choice. Besides, TDLibWrapper is already pretty big, I would be careful about dragging new stuff in there.

src/chatlistmodel.cpp
jgibbon: Huh, the warnings I was talking about may already have been resolved by changing other things in between. Didn't get any after just putting it in. Nice.

jgibbon:
Yup, sounds reasonable. 

What are your general thoughts on the rest of this PR? 
Would it make sense to go on with the other changes? If so, I'd restore the cache to the model and try to rip it out of the wrapper. 

BTW: There's yet another Map for secret chats there – but I haven't looked at it much. I'd leave that be for the moment if it's not too badly intertwined.

Thanks for having a look!

monich:
Storing QVariantMap in a parsed form definitely makes sense. You access the field directly instead of going though a QVariantMap lookup. You don't need to store the keys (essentially field offset becomes the key). Some types (especially the primitive ones, like bools, ints or longlongs) are A LOT more efficient to store in a parsed form vs wrapping them into a QVariant. And so on.

jgibbon:
OK, thanks. 
I think I'll do that part "again" in a dedicated PR – that should be a lot easier to test. 
We can then think about all the ever growing caches of chats (and other data like users) separately. 

Closing this but leaving the branch up for reference.

PR #340, opened 2021-01-26
User profile and privacy settings

### LINKED ISSUE #155: Add privacy settings/online state
Right now, AFAICT, Fernschreiber does not tell Telegram if it it used, which is a good default as long as the privacy settings aren't implemented.

The Settings in question are userPrivacySettingAllowCalls, userPrivacySettingAllowChatInvites, userPrivacySettingAllowFindingByPhoneNumber, userPrivacySettingAllowPeerToPeerCalls, userPrivacySettingShowLinkInForwardedMessages, userPrivacySettingShowPhoneNumber, userPrivacySettingShowProfilePhoto, and userPrivacySettingShowStatus.

I think it would also be a good idea to automatically set privacy-aware settings after setting up an account.

Wunderfitz:
This change brings user profile and privacy settings to the settings page. It fixes #155 (intentionally omits certain options that Fernschreiber doesn't support yet (such as calls).
It includes:
- First Name
- Last Name
- Username
- Profile Pictures
- Relevant privacy settings


qml/pages/OverviewPage.qml
```diff
@@ -92,6 +92,12 @@ Page {
             tdLibWrapper.getRecentStickers();
             tdLibWrapper.getInstalledStickerSets();
             tdLibWrapper.getContacts();
+            tdLibWrapper.getUserPrivacySettingRules(TelegramAPI.SettingAllowChatInvites);
```
jgibbon, line 95: It's a bit like the requests above, but: Do we actually need them at this point? Getting them on demand inside getUserPrivacySettingRule if they're not filled would be pretty nice. The "empty" result would then be updated as soon as the real result arrives, anyway.

More aggressive alternative approach: Do we even need to cache them at all?

jgibbon:
Nice initial implementation, good work! 
One small remark above, also in the future, the rule exception features and blocked users could be useful. 
But this is pretty awesome as it is!
Thumbs up!

qml/pages/OverviewPage.qml
Wunderfitz: :) I'm probably over-pessimistic, I simply want to have the data ready when the user opens up the settings page without values flickering around when the page is opened. There's always a trade-off and - you know me - I tend to cache more....

Wunderfitz:
Thank you very much for the review. Yes, the exception features are also interesting, but I'd postpone them...

PR #347, opened 2021-01-29
Don't allow ContextMenu to be wider than its ComboBox

monich:
That looks ridiculous:

Screenshot_20210130_001

and this I think is much better:

Screenshot_20210130_002

Honestly speaking, I feel the urge to redo this new User Profile section, but let's start with smaller and more obvious things.


Wunderfitz:
Oh cool, wouldn't have expected that the solution is that easy. Thank you very much!

And feel free to redo the section, I was simply happy to find some time to do it at all...

PR #348, opened 2021-01-30
Don't display default and superlowres images at the same time

monich:
It looked like this:

a

Also, I don't think asynchronous loading mode makes sense for these minithumbnail images. That may actually introduce even greater overhead.


jgibbon:
Displaying the background image as well was meant as an indicator that this isn't the final state - mostly for people with really slow connections. But I'm not really opposed to removing it.
I agree with the synchronous image.

monich:
I think it's pretty obvious that this blurry thing isn't the final state 🙂

Wunderfitz:
OK, let's do it like that then... Thanks for the changes!

PR #349, opened 2021-01-30
Made notification sound configurable

monich:


Wunderfitz:
Not really sure if we should do this. With the recent change, Fernschreiber uses the "Chat" setting from the overall device settings. If this is set to "No Sound", Fernschreiber also remains silent (that's actually my default setting ;)) - if a sound is configured this one is played properly. 

If we introduce this additional setting here, we would need to somehow override the global setting, especially if the global one set to "No Sound". Just tested it on my device - of course it remains silent even if your new setting is switched on. Right now I'd prefer to keep it as is - using the OS-wide setting for chats, but I'm open to a discussion here...

monich:
That's slightly confusing indeed, if we can only disable sounds and not enable them. I'll turn this option into "Disable sounds" then, is that OK?

Wunderfitz:
Yep, that would be fine. But please also add an explanatory text that this one overrides the SFOS general settings. Something like "This setting overrides the Sailfish OS notification settings for chats and disables notification sounds for Fernschreiber completely" would be appropriate IMHO.

monich:
How about this:

✅ **Enable notification sounds**
_When sounds are enabled, Fernschreiber would use Sailfish OS notification sound for chats, which can be configured in system settings._

That makes it crystal clear that if you select "No sounds" for chats in system settings, you won't get any sounds, even if you enable sounds in Fernschreiber.

I like it when checkboxes enable things rather than disable. It simplifies the logical connection between the UI option and its effect, checked = enabled, unchecked = disabled. When unchecked means "not disabled", it kind of requires additional mental effort to digest double negation and realize that "not disabled" actually means "enabled".

Wunderfitz:
Yep, agree with you on the semantics of a checkbox. Please be sure to include a proper explanation - then we can do it.

monich:
Rebased and added description for the TextSwitch.

Wunderfitz:
Great, thanks!

PR #350, opened 2021-01-30
Tweak cover text size and alignment

monich:
Before:

Screenshot_20210130_005

and after:

Screenshot_20210130_004

What do you think?

### DIFF
--- a/qml/pages/CoverPage.qml
+++ b/qml/pages/CoverPage.qml
@@ -122,7 +122,6 @@ CoverBackground {
         visible: coverPage.authenticated
         Row {
             width: parent.width
-            spacing: Theme.paddingMedium
             Text {
                 id: unreadMessagesCountText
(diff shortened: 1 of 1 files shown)


Wunderfitz:
I can't help myself, but I'm still in favor of keeping it vertically centered. To me bottom-aligned looks weird, especially when there aren't that many messages, see the screenshots for instance. However, we can talk about a smaller font size for the text...

image
image

Wunderfitz:
Sorry, I actually wanted to close #351... But as said in the chat - I really like to keep the texts vertically centered. Thus, closing this one would be the next consequence. But feel free to reopen it in case you want to continue here.

monich:
Never mind, it was just a suggestion!

Wunderfitz:
But a good one - I would have never noticed the font size thingy...

PR #354, opened 2021-02-01
Explicitly set image-path notification hint to empty string

monich:
This stops home screen (on a really really really fresh build of Sailfish OS) from using app icon as a default:

Screenshot_20210201_001

This `image-path` thing (or `image-data`) may actually be used at some point (e.g. for passing minithumbnail if it's available) but for now let's just pass an empty string, it doesn't cost much. The same icon all over the place just looks silly and eats up horizontal space.

With an empty `image-path` the second icon disappears:

Screenshot_20210201_002


jgibbon:
We could test this if you'd put in a good word for us to be invited to cbeta. ;)

As i've tried before and a previous community manager forgot he wanted to get in touch, I think we could just merge this.

Wunderfitz:
Yes, let's merge this. :)

Issue #301, opened 2021-01-09
Inconsistency in message delivered and read indication


lal883:
I was trying out the secret chat feature today and noticed this.

In a secret chat I was messaging a person who was offline for some time and all messages I sent had a green tick indicating they have read those. But If I swipe back to the chat selection page (start page) the same message has a blue tick in the over view, which is actually correct. They have not read the message, but probably the messages were delivered. Have attached a couple of screenshots below.

Chat page (green tick, should be blue tick here I believe)
![image](https://user-images.githubusercontent.com/22425235/104089587-2b4d7280-5296-11eb-8950-815c4d7e8d6e.png)

Chat selection/main page (blue tick, correct indication)
![image](https://user-images.githubusercontent.com/22425235/104089510-be39dd00-5295-11eb-9993-74f8781716d8.png)

Issue #271, opened 2021-01-03
Open Chat Scroll Position: Own unread Messages


jgibbon:
I've found a specific variant of the closed and pretty general #1.

To reproduce it, you just have to spam a bit without your messages being read by the other side. 
When re-opening the chat, it will only be scrolled down to the last read message, not your most recent. 

At least this happened to me reproducible in both a private chat and a (basic) group chat.

jgibbon:
This may or may not be the thing mentioned in the comments of #243, but I've understood #243 as a different thing/feature request.

yusssufff:
The overall #243 issue is the sort of "random landing" when opening a chat...
Usually landing pretty close to last message but often a few messages above.... 
S when opening a new chat first thing is often the need to scroll down.
At least on my device (xa2).
I couldnt find an exact "logic" for this "almost last message landing".... Sometimes it is last read, sometimes last recieved...
Interestingly Depecher always lands on last message for example.

Issue #243, opened 2020-12-16
landing on last message directly when opening a chat


yusssufff:
It would be great to land on the last message directly when opening a chat (it is already better with latest versions of Fernschreiber) so not to have to scroll down immediately after opening a chat...

monich:
First unread, I would say. And it actually seems to land pretty close to it already (like, last read message, you only need to scroll down a little bit).

Wunderfitz:
I'm honestly really happy with the status quo and if there is a change in behavior, I'd like to have it configurable. So, please don't expect any changes here from my end in the near future, but - as always - contributions are welcome! :)

yusssufff:
Well on my device, when i open a chat it lands as Monich funnily says "pretty close" to the last message:).
But there isnt really folling a rule that i could understand. Sometimes it lands on last message sent, sometimes on the last recieved, sometimes a bit more up, sometimes in the middle of a message...
Truly said there isnt much to scroll down (It used to be much more problematic a lot of couple of Fernschreiber's versions ago), but its just a repetitive  move of scroll down each time of opening a chat....

PR #359, opened 2021-02-02
Improved chat message index handling

### LINKED ISSUE #235: Unread count sometimes not cleared even when I answered to that chat
Sometime the unread count does not get cleared, even when I answered to a chat (i.e. spend "enough" time at the latest message). The last x messages (including my answers) will then be highlighted as unread. 
The unread counter gets cleared, when I reenter the chat, flick to end of conversation and wait 2 sec. 

I can not (yet) reliably recreate this bug, it just happens sometimes, both in single person conversations, but also in group chats. I have not tested it with secret chats.

I am using Fernschreiber 0.51, Sailfish OS 3.4.0.24 on an XA2 (single sim).  
I also have telegram running on my laptop.

Wunderfitz:
This change tries to improve the chat message index handling, this one fixes #235, #243, #271 and #301 (hopefully):
The errors were:
- Unread count wasn't set to zero in case you sent out a new message without scrolling down to actually 'read' it actively (#235)
- Try to display the first unread message after opening a chat (#243)
- Own messages were marked as unread after receiving a new message (#271)
- Sent messages were marked as read though they haven't been read yet (#301)

Happy testing!

### DIFF
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -443,7 +443,6 @@ Page {
                 chatModel.initialize(chatInformation);
 
                 pageStack.pushAttached(Qt.resolvedUrl("ChatInformationPage.qml"), { "chatInformation" : chatInformation, "privateChatUserInformation": chatPartnerInformation, "groupInformation": chatGroupInformation, "chatOnlineMemberCount": chatOnlineMemberCount});
-                chatPage.isInitialized = true;
 
                 if(doSendBotStartMessage) {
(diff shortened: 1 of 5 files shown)


jgibbon:
Even though I didn't have time to test all eventualities I could come up with (nor had the time to understand each change in detail), this is without a doubt a really good improvement! Excellent work!

Thumbs up for merging.

Wunderfitz:
Then let's merge it, thanks!

PR #361, opened 2021-02-03
Fix click signal forwarding for extra content items

jgibbon:
From the dev chat in Telegram:
> I guess its an already known thing that: when trying to display full screen a picture only clicking left side of it opens it?

I think it wasn't. ;)


Wunderfitz:
:) Thx!

PR #363, opened 2021-02-03
"Fix" Videos/VideoPage

jgibbon:
This is a _really_ quick fix to get the video page working before the next release.

Because there was so much stuff (like multiple control button panels) inside MessageVideo and VideoPage, I basically removed a lot of it, shuffled the rest around to get things visible above the video and quickly introduced TDLibFile.
Layout is not exactly the same as before, but looks kind of alike.

I'm not 100% satisfied with this. For example, the Video Loader had to go, as well, to save time… Also it's not a real improvement on the previous layout.
But perhaps it's enough to further build upon. ;)

### DIFF
--- a/qml/components/messageContent/MessageVideo.qml
+++ b/qml/components/messageContent/MessageVideo.qml
@@ -18,6 +18,7 @@
 */
 import QtQuick 2.6
 import Sailfish.Silica 1.0
+import WerkWolf.Fernschreiber 1.0
 import QtMultimedia 5.6
 import "../"
 import "../../js/functions.js" as Functions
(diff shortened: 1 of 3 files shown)


monich:
I would suggest to reserve names that start with `TDLib` for native objects. That of course, can wait and isn't related to this PR.

qml/components/messageContent/MessageVideo.qml
```diff
@@ -26,18 +27,26 @@ import "../../js/debug.js" as Debug
...
 
-    property var videoData:  ( rawMessage.content['@type'] === "messageVideo" ) ?  rawMessage.content.video : ( ( rawMessage.content['@type'] === "messageAnimation" ) ? rawMessage.content.animation : rawMessage.content.video_note )
-    property string videoUrl;
-    property int previewFileId;
-    property int videoFileId;
-    property bool isVideoNote : false;
-    property bool fullscreen : false;
+    property bool isVideoNote : rawMessage.content['@type'] === "messageVideoNote";
+
+    property var contentBase: ( rawMessage.content['@type'] === "messageVideo" ) ?  rawMessage.content.video : ( ( rawMessage.content['@type'] === "messageAnimation" ) ? rawMessage.content.animation : rawMessage.content.video_note )
```
monich, line 32: Not that it really matters, but I would pass `content_type` role as a property to `MessageContentBase` to avoid these `content['@type']` lookups.

Wunderfitz:
Not really sure... Had some issues in fullscreen after the video is finished. Moreover, the progress bar remains there after the video has finished and I had some display issues with landscape videos if finished in fullscreen... Will check a bit more later - also maybe I can fix the original version, I had several reasons for the original version. ;)

PR #370, opened 2021-02-07
Use different badge colors for muted chats

monich:
Unmuted:

Screenshot_20210207_001

Screenshot_20210207_003

Muted:

Screenshot_20210207_002

Screenshot_20210207_004

Linux desktop client is doing something like that too. What do you think?


Wunderfitz:
Yep, makes sense. Thanks for that!

PR #373, opened 2021-02-07
Hardcode line break in "unread messages"

monich:
To avoid cover layout change when number of unread messages goes 9 => 10:

Screenshot_20210207_006
Screenshot_20210207_005

I like it better when there's line break even with a single digit:

Screenshot_20210207_007

What do you think?


monich:
And of course, since wrapping depends on the font size etc., on some devices (e.g. Jolla tablet) it's always wrapped even without hardcoding the line break. These screenshots were taken on Sony XA2.

Wunderfitz:
Yeah, why not... :)

PR #380, opened 2021-02-15
Update harbour-fernschreiber-ru.ts

arustg:


Wunderfitz:
Thanks! But please remove the "unfinished" attributes for all the translations you want to be used in the productive version. If the "unfinished" attribute is still there, these ones are usually ignored by the localization framework.

arustg:
Done!

PR #382, opened 2021-02-15
Introduce message options drawer

### LINKED ISSUE #216: Too many items in the message context menu
All 6 items (Reply, Edit, Copy, Select, Pin, Delete) don't fit into the screen in landscape on most devices. Even 5 is one too many, I would say - the menu fills the entire chat area but at least it doesn't get clipped. Sixth item does get clipped.

Wunderfitz:
This change introduces a message option drawer that may hold all message options that would otherwise make the context menu become too big. With this change
- the message context menu should have 4 items maximum (one of them is "More Options...")
- this "More Options..." option opens the drawer (at the bottom on portrait, to the right on landscape) which contains the other options
Consequently, this fixes #216 :)

### DIFF
--- a/qml/pages/ChatPage.qml
+++ b/qml/pages/ChatPage.qml
@@ -335,6 +335,18 @@ Page {
 
     }
 
+    function startForwardingMessages(messages) {
+        var ids = Functions.getMessagesArrayIds(messages);
+        var neededPermissions = Functions.getMessagesNeededForwardPermissions(messages);
+        var chatId = chatInformation.id;
+        pageStack.push(Qt.resolvedUrl("../pages/ChatSelectionPage.qml"), {
+            myUserId: chatPage.myUserId,
+            headerDescription: qsTr("Forward %Ln messages", "dialog header", ids.length),
+            payload: {fromChatId: chatId, messageIds:ids, neededPermissions: neededPermissions},
+            state: "forwardMessages"
+        });
+    }
+
     function forwardMessages(fromChatId, messageIds) {
         forwardMessagesTimer.fromChatId = fromChatId;
         forwardMessagesTimer.messageIds = messageIds;
@@ -696,6 +708,14 @@ Page {
                 visible: true
                 name: qsTr("Copy Message to Clipboard")
                 action: function () { Clipboard.text = Functions.getMessageText(messageOptionsDrawer.myMessage, true, messageOptionsDrawer.userInformation.id, true); }
+            },
+            NamedAction {
+                visible: messageOptionsDrawer.myMessage.can_be_forwarded
+                name: qsTr("Forward Message")
+                action: function () {
+                    var messagesToForward = [ messageOptionsDrawer.myMessage ];
+                    startForwardingMessages(messagesToForward);
+                }
             },
             NamedAction {
                 visible: canPinMessages()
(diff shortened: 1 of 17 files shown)


Wunderfitz:
The changelog looks crazy, but in the end, I only surrounded the main `SilicaFlickable` of the `ChatPage` with the new `Drawer` and added the other options...

qml/pages/ChatPage.qml
```diff
@@ -681,1172 +681,1307 @@ Page {
...
+                width: parent.width
+                height: parent.height - drawerHeaderRow.height
+                contentHeight: drawerContentColumn.height
+                clip: true
+                Column {
+                    id: drawerContentColumn
+                    width: parent.width
+                    Repeater {
+                        model: messageOptionsDrawer.additionalItemsModel
+                        delegate: BackgroundItem {
```
jgibbon, line 742: One could possibly put these BackgroundItems in a separate qml because they are basically the same thing, this may even reduce the parsing/compiling time a tiny bit. 
Also (you probably tried already): Doesn't a ContextMenu with MenuItems work here instead of the Column?

qml/pages/ChatPage.qml
```diff
@@ -681,1172 +681,1307 @@ Page {
...
-                id: closeSecretChatMenuItem
-                visible: chatPage.isSecretChat && chatPage.secretChatDetails.state["@type"] !== "secretChatStateClosed"
-                onClicked: {
-                    var secretChatId = chatPage.secretChatDetails.id;
-                    Remorse.popupAction(chatPage, qsTr("Closing chat"), function() { tdLibWrapper.closeSecretChat(secretChatId) });
-                }
-                text: qsTr("Close Chat")
-            }
+        dock: chatPage.isPortrait ? Dock.Bottom : Dock.Right
+        backgroundSize: dock == Dock.Left || dock == Dock.Right ? width / 2 : height / 3
```
monich, line 701: Spent some time trying to figure how `dock` could become `Dock.Left` but no avail. You probably meant this:
```js
backgroundSize: chatPage.isPortrait ? height / 3 : width / 2
```

qml/pages/ChatPage.qml
monich: > Also (you probably tried already): Doesn't a ContextMenu with MenuItems work here instead of the Column?

Or a SilicaListView with MenuItems in it.

jgibbon: Yeah, that would probably make even more sense ;)

qml/pages/ChatPage.qml
```diff
@@ -681,1172 +681,1307 @@ Page {
...
+                            anchors.verticalCenter: parent.verticalCenter
+                            anchors.horizontalCenter: parent.horizontalCenter
+                            horizontalAlignment: Text.AlignHCenter
+                        }
+                    }
+                    BackgroundItem {
+                        width: parent.width
+                        visible: messageOptionsDrawer.myMessage.can_be_deleted_for_all_users || (messageOptionsDrawer.myMessage.can_be_deleted_only_for_self && messageOptionsDrawer.myMessage.chat_id === chatPage.myUserId)
+                        onClicked: {
+                            var chatId = page.chatInformation.id;
```
jgibbon, line 796:
```suggestion
                            var chatId = chatPage.chatInformation.id;
```

Wunderfitz: Absolutely correct...

qml/pages/ChatPage.qml
Wunderfitz: Oh, I simply copied the content from `Dock` and changed the values. Of course, your proposal makes much more sense...

qml/pages/ChatPage.qml
Wunderfitz: Yesterday, I didn't make it to join two different lists of `NamedAction`s, but today it worked with probably the most embarrassing piece of JS code from the last century - but it seems to work.

Wunderfitz:
Merging this one now as people seem to be happy. Thanks for testing and your feedback - really appreciated! :)

PR #390, opened 2021-02-20
Don't have on-screen keyboard and drawer visible simultaneously

monich:
Also, removed unnecessary `closeRequested()` signal.

Generally, my feeling is that since this drawer thing is sort of an extension to the context menu, it should behave more like a context menu, i.e. have a stronger visual association with the item it belongs to.

### DIFF
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -80,9 +80,11 @@ ListItem {
     }
 
     Connections {
-        target: messageOptionsDrawer
-        onCloseRequested: {
-            messageListItem.additionalOptionsOpened = false;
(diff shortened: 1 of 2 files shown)


Wunderfitz (changes requested):
I'm sorry, but there is a reason for this indirection. If you remove the `additionalOptionsOpened` like this, all list items that are currently loaded will have `highlighted = true` as you can see in the screenshot. Of course, the other change (drawer and keyboard must not be visible at the same time) is totally fine and I really appreciate it.

image

monich:
Yeah, strange that I haven't noticed it. Good catch.

monich:
Fixed and rebased.

Wunderfitz:
...and merged! Thanks for the changes!

PR #391, opened 2021-02-20
Fix MessageSticker highlight and click behavior

monich:
The `highlighted` property has been moved to `MessageContentBase` and the mouse area was stealing mouse events from the parent.


Wunderfitz:
Good catch, thank you very much!

PR #399, opened 2021-02-23
Updated spanish transtation

GNUuser:
Can you explain me which  feature is Ln: 1705

### DIFF
--- a/translations/harbour-fernschreiber-es.ts
+++ b/translations/harbour-fernschreiber-es.ts
@@ -301,7 +301,7 @@
     </message>
     <message>
         <source>No message in this chat.</source>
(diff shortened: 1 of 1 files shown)


Wunderfitz:
Sure. By default, the app waits a second until a message that is on the screen is marked as read. If you switch this feature off, a message will immediately marked as read once it's on the screen (and you're not scrolling at the moment).

Wunderfitz:
Gracias

GNUuser:
Ok i checking......

GNUuser:
Very confusing this feature, and may be not useful. But i will how fix this confusing..

PR #401, opened 2021-02-25
Don't show Active Sessions button when not authenticated

monich:
Also, expose `authorizationState` as `TDLibWrapper` property to simplify the logic a bit and slightly reduce the amount of QML code.


Wunderfitz:
Thanks, completely agree, that makes sense.

PR #411, opened 2021-03-26
Handle more cases when the drawer should be closed

monich:
1. Other item is pressed
2. Any item is long-pressed, including the selected one
3. ChatPage is getting deactivated


Wunderfitz:
Thanks, sorry for the late reply

PR #414, opened 2021-05-01
Cmake

simonschmeisser:
This is an attempt to build fernschreiber all in one, without manually checking out tdlib and copying the lib around

I needed to update rlottie as it has some issues with being included as a CMake subfolder. This was fixed in a newer commit. Another issue with the include path not being exported is still open, see the PR herehttps://github.com/Samsung/rlottie/pull/476

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,7 @@
 [submodule "rlottie"]
 	path = rlottie
 	url = https://github.com/Samsung/rlottie.git
+[submodule "td-src"]
+	path = td-src
+	url = https://github.com/Wunderfitz/td.git
+	branch = fernschreiber
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,116 @@
+project(harbour-fernschreiber CXX)
+cmake_minimum_required(VERSION 3.5)
+
+find_package(Qt5 COMPONENTS Core Network Qml Gui Quick Sql Positioning DBus Multimedia REQUIRED)
+
+include(FindPkgConfig)
+pkg_search_module(SAILFISH sailfishapp REQUIRED)
+
+pkg_search_module(NEMONOTIFICATIONS-QT5 nemonotifications-qt5 REQUIRED IMPORTED_TARGET)
+
+find_package(ZLIB REQUIRED)
+
+set(CMAKE_AUTOMOC ON)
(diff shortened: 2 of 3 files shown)


Wunderfitz:
Thanks for the contribution! However, I have to admit that I have almost zero knowledge about CMake. As I'd like to focus on development and not on maintaining the build files, I'm not sure if I could maintain that... I also have no idea how it fits together smoothly with the existing build files - especially if things are updated or new files come in. I assume double maintenance would be required. 

I'd probably be willing to include TDLib as a submodule (though it's updated rarely and each time it needs to be built I'd need to give the build environment more memory), but I'd rather have it my normal toolchain than in a new one. 

Consequently, I'd have some questions:
- Would you be able to maintain the CMake files permanently?
- What's the benefit of CMake that we can't have with the existing toolchain? If it's only to get TDLib built with the app, I have to admit that I don't see a huge benefit. The files are updated maybe once or twice a year and can be included with other means.
- Could things be automated that the CMake files are updated automatically when I change the project file(s) in Qt Creator?
- Did you test if the yaml file is changed in Qt Creator, are your changes still there afterwards or are they overwritten?

And finally: In case we agree to merge the changes, I assume the README needs some update to mention the build options with CMake... ;)

simonschmeisser:
On the contrary I have no experience with qmake.

My main motivation was that I felt intimidated by the build  instructions in the readme and thought it should be possible with just one command. Now it's also easier to add github actions to build this automatically on each commit/merge request.
QtCreator unfortunately cannot add new files automatically to the CMakeLists.txt, that needs to be done manually. But there could be a red cross right next to the commit telling you that you missed adding some files ...

I just bought a XA2 because pine phone's camera just didn't cut it and will be around for some while!

Which yaml file?

I will open further PRs for features based on this but can rebase them before merging so this can also sit around for some while, no worries

Wunderfitz:
OK, in this case I'd like to keep it open until the most relevant synchronization issues are done. Feel free to keep on adding stuff here. I'll also give it a try when I have more time. 

About the YAML, I'm referring to this one: https://github.com/Wunderfitz/harbour-fernschreiber/blob/master/rpm/harbour-fernschreiber.yaml Please see https://github.com/mer-tools/spectacle for details about the underlying toolset.

simonschmeisser:
Sure, no need to rush anything and thanks for pointing me to spectacle.

PR #416, opened 2021-05-04
Fix new line in TextInput when using Send By Enter

### LINKED ISSUE #406: Text input inputs newline when you hit enter
When you have selected in options "send message with enter" and you press enter with the cursor between lines in a message, then the message is sent but with a line break where cursor position
photo5911117761424242177
photo5911117761424242178

santhoshmanikandan:
Fixed the line feed that gets added to message text if the cursor is placed anywhere in-between the message text.

Fixes #406


Wunderfitz:
Thx

PR #419, opened 2021-05-22
Speed up Sticker Picker

jgibbon:
Just a quick fix to
- speed up the picker for users with many sets by removing those repeaters
- add some background areas for clickable areas
- add a transition for opening the sticker sets (just for fun)

There are a few things that could be further optimized, but it's already quite a bit faster to get to an "interactable" state for me.

cheers

### DIFF
--- a/qml/components/StickerPicker.qml
+++ b/qml/components/StickerPicker.qml
@@ -165,7 +165,6 @@ Item {
                             verticalCenter: parent.verticalCenter
                             margins: Theme.paddingSmall
                         }
-                        maximumLineCount: 1
                         truncationMode: TruncationMode.Fade
                         text: modelData.title
                     }
--- a/qml/components/StickerPicker.qml
+++ b/qml/components/StickerPicker.qml
@@ -44,6 +44,30 @@ Item {
             installedStickerSets = stickerManager.getInstalledStickerSets();
         }
     }
+    Component {
+        id: stickerComponent
+        BackgroundItem {
+           id: stickerSetItem
+           width: Theme.itemSizeExtraLarge
+           height: Theme.itemSizeExtraLarge
+
+           onClicked: stickerPickerOverlayItem.stickerPicked(modelData.sticker.remote.id)
+
+           TDLibThumbnail {
+               thumbnail: modelData.thumbnail
+               anchors.fill: parent
+               highlighted: stickerSetItem.highlighted
+           }
+
+           Label {
+               font.pixelSize: Theme.fontSizeSmall
+               anchors.right: parent.right
+               anchors.bottom: parent.bottom
+               text: Emoji.emojify(modelData.emoji, font.pixelSize)
+           }
+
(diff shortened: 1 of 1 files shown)


qml/components/StickerPicker.qml
```diff
@@ -54,31 +78,28 @@ Item {
...
 
-        Column {
-            id: stickerPickerColumn
-            spacing: Theme.paddingMedium
-            width: stickerPickerFlickable.width
+        model: stickerPickerOverlayItem.installedStickerSets
+
+        header: Column {
+            width: stickerPickerListView.width
+            height: childrenRect.height
```
monich, line 90: This sort of binding looks strange for a Column. Doesn't it automatically set its own height to the total height of its children, laid out vertically?

qml/components/StickerPicker.qml
```diff
@@ -87,164 +108,133 @@ Item {
...
+                        font.pixelSize: Theme.fontSizeLarge
+                        font.bold: true
+
+                        anchors {
+                            left: stickerSetThumbnail.right
+                            right: expandSetButton.left
+                            verticalCenter: parent.verticalCenter
+                            margins: Theme.paddingSmall
+                        }
+                        maximumLineCount: 1
```
monich, line 168: I don't think this makes any difference when wrapping if off (which is the default)

jgibbon: Yup, I just copy/pasted most stuff from the previous implementation. Good catch!

qml/components/StickerPicker.qml
jgibbon: That's what I thought, yes. 
But it does not, for some reason unknown to me. 
If height isn't defined, we only get the height of the Label.

monich: I have a feeling that it's somehow related to the `visible` attribute. Would it make sense to do `height: (count > 0) ? Theme.itemSizeExtraLarge : 0` instead? That would even allow to animate `recentStickersGridView` height, if needed.

jgibbon: It would make more sense to do a mix of that and the existing code for the whole column. This would also hide the Label if there are no recent stickers. 
Alternatively, display another Label describing there's nothing to see here.
I don't think a transition/animation is needed here, the recently used stickers don't really update "in place".

```
Column {
 visible: recentStickersGridView.visible
 height: visible? childrenRect.height : 0
 […]
}
```
or something in that regard. Thoughts?

Wunderfitz: I agree that we don't need an animation here. But the childrenRect also looks strange to me. I changed it to use a calculated height for all children including padding if the count is greater than zero - along with some other optimizations.

Wunderfitz:
Thank you very much for the work and also the discussion. The new picker is really nice - I'll merge it now. If further optimizations are needed, we can adjust them still later. :)

PR #421, opened 2021-05-26
Add Github action

jgibbon:
As far as I've seen, it works well – but I've lacked the time to test the resulting packages thoroughly, just quickly installed/ran on my dev XA2.

Minimal documentation of what to do and how to use it is inside the README.md – basically, it tries to build after each push to master (we could limit that in the future). After adding the repository secrets, these builds should even work and produce an artifact .zip. Pushing a tag like "v0.123.9" creates a github release, "pre-0.123.9" creates a release marked as "pre release". See https://github.com/jgibbon/harbour-fernschreiber/releases.

Aarch64 builds on a 4.0 target to avoid different errors with both 3.4 and the brand new 4.1 target. The other architectures use 3.4.

cheers!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,105 @@
+name: Fernschreiber build
+
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - '*'
+  # to prevent secrets leaking,
+  # we don't build on PRs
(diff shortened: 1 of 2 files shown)


Wunderfitz:
Thank you very much!

PR #422, opened 2021-05-26
Delay link/notification handling until chatListCreated

### LINKED ISSUE #412: Opening a Telegram channel displays at least one error message before actually working
I use Sailfish 4.0.1.48 with the latest Fernschreiber. When I've loaded https://t.me/aurorasupport/ in the Sailfish browser and tap "View in Telegram", I get this error: "Initialization parameters are needed: call setTdlibParameters first". When I go back to the browser and tap again, it works. Although there was one occasion, with another group I cannot remember, when the second time I got a second error about some database and checking encryption. The third time finally worked. None of these issues occur with the desktop Telegram client for macOS only.

jgibbon:
(hopefully) fixes #412

`openUrl` could have been earlier but I decided to do both at roughly the same time because they are pretty similar.

I have included messageId in case we want to use it in the future (also it's `onPleaseOpenMessage`), but it's not really used otherwise.

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -36,6 +36,10 @@ Page {
     property int ownUserId;
     property bool chatListCreated: false;
 
+    // link handler:
+    property string urlToOpen: null;
+    property var chatToOpen: null; //null or [chatId, messageId]
+
     onStatusChanged: {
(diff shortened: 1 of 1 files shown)


Wunderfitz:
Thank you very much!

PR #429, opened 2021-06-02
Switch to generic booster

monich:
That seems to fix a mysterious aarch64 crash with some animated stickers.


jgibbon:
Without a device, I can only look for an obvious negative impact on other things.

While I thought I had seen a weird behaviour with the notification/dbus interaction while trying this (independently from this PR), it turned out that was some other things I tried simultaneously (which is always a good idea, isn't it).

Weird situation, but a really nice find! Hopefully you guys find the booster culprit at some point.

:+1: from me.

PR #440, opened 2021-09-18
Animate accordion image rotation

monich:
Rotation angles are changed to make it behave more like event view buttons (arrow pointing inside the screen during animation)


Wunderfitz:
Sorry, obviously missed that PR... Cool change, thanks for that!

PR #445, opened 2021-10-29
fix for 'sent' traduction

nbourdais:
'sent' was translated in present form. This PR fixes it with past form


Wunderfitz:
Merci!

PR #449, opened 2021-12-04
Use larger font for settings section headers

monich:
Small font in settings section header is bothering me 🙂

Screenshot_20211205_001 Screenshot_20211205_002

Small size on the left, default (proposed) size on the right. What do you think?

### DIFF
--- a/qml/components/settingsPage/AccordionItem.qml
+++ b/qml/components/settingsPage/AccordionItem.qml
@@ -78,7 +78,6 @@ Item {
             }
             horizontalAlignment: Text.AlignRight
             truncationMode: TruncationMode.Fade
-            font.pixelSize: Theme.fontSizeSmall
             color: button.highlighted ? Theme.highlightColor : Theme.primaryColor
             textFormat: Text.PlainText
         }


mschilli87:
TBH, I cannot make out any difference between the two versions. :shrug:

monich:
> _edit_: After looking at the code change I think I can tell the menu options to be a bit larger on the right side but it is so subtle that to me it wouldn't make a difference if this got merged or not.

The right picture with only 3 different fonts looks significantly calmer to me, the left one has 4 difference fonts, that's one too many... Section headers already differ from e.g. button labels by alignment, background and the arrow icon, there's no need to emphasize the difference even further. Too much variety, that's what was bothering me I guess. That's how I can explain my reaction.

And yeah, many people wouldn't tell the difference. Some are more sensitive to this sort of visual details, some less. But if some people like the change and some don't feel any difference, then the overall effect would be positive. If there are no dislikes, of course. But why would anyone dislike it 🙂

Wunderfitz:
We take it :D

PR #450, opened 2021-12-05
Hide "Extra options..." context menu item if it's not needed

monich:
If all context actions fit nicely into the context menu limit (5 items in portrait, 4 items in landscape) it makes no sense to show this "Extra options..." thing.


Wunderfitz:
Thanks, didn't have time to test it completely, but I trust you! :) And if we find issues, there will certainly be a bugfix release ;)

PR #455, opened 2021-12-06
Attempt to optimize sponsored message handling

monich:
By reducing number of `QVariantMap` lookups and various run-time conversions to/from `QString`. At expense of allocating even more memory for `messageType`🙁

It's hard to tell whether `messageType` allocation vs `QVariantMap` lookup trade-off really makes sense. Static `TYPE_SPONSORED_MESSAGE` and use of already existing `_TYPE` key definitely do make sense, though.


monich:
Pushed another commit which adds an important option

Wunderfitz:
Thanks for that!

PR #460, opened 2021-12-07
Describe emoji with emoji

monich:
And stickers too. In other words, replace this:

Screenshot_20211207_002

with this:

Screenshot_20211207_001

What do you think?

### DIFF
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -47,9 +47,9 @@ function getMessageText(message, simple, currentUserId, ignoreEntities) {
             return enhanceMessageText(message.content.text, ignoreEntities);
         }
     case 'messageSticker':
-        return simple ? qsTr("Sticker: %1").arg(message.content.sticker.emoji) : "";
+        return simple ? message.content.sticker.emoji : ""
     case 'messageAnimatedEmoji':
-        return simple ? qsTr("Animated Emoji: %1").arg(message.content.animated_emoji.sticker.emoji) : "";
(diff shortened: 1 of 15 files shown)


qml/js/functions.js
```diff
@@ -47,9 +47,9 @@ function getMessageText(message, simple, currentUserId, ignoreEntities) {
             return enhanceMessageText(message.content.text, ignoreEntities);
         }
     case 'messageSticker':
-        return simple ? qsTr("Sticker: %1").arg(message.content.sticker.emoji) : "";
+        return message.content.sticker.emoji
     case 'messageAnimatedEmoji':
-        return simple ? qsTr("Animated Emoji: %1").arg(message.content.animated_emoji.sticker.emoji) : "";
+        return message.content.animated_emoji.sticker.emoji
```
Wunderfitz, line 52: This change causes sticker emojis and animated emojis to duplicate in chats. You should still check for the `simple` parameter.

Wunderfitz (changes requested):
Also thought about that after the 0.11 release. ;) So in general, I support the change, but in functions.js, you should still check for the `simple` parameter.

qml/js/functions.js
monich: Right! Good catch.

monich: Fixed!

monich:
And I'll probably add an option to show animated emoji as a sticker rather than an image. But that would be a separate patch.

Wunderfitz (approved):
Thanks for that! :)

PR #462, opened 2021-12-08
Apply last remaining sticker option to animated emoji

monich:
In all other respects their behavior is already equivalent. Providing a whole separate set of options for animated emoji feels like an overkill to me.


Wunderfitz:
Sure thing, thanks!

PR #464, opened 2021-12-09
Improve chat search performance

monich:
Simple text search performs significantly better that regular expressions.

Disconnecting search filter model from the source when search is off should be good for performance because the filter model won't have to unnecessarily react to the source model changes.


Wunderfitz:
Thanks for that! I probably wouldn't have noticed that as I don't have that many chats...

PR #465, opened 2021-12-09
Hide appearance settings that are not applicable

monich:


Wunderfitz:
Good catch, thanks! :)

PR #466, opened 2021-12-10
Optimized forwarding

monich:
It was a very strange experience. Now it's more like what an average user would expect 🙂

I hope I correctly reproduced in native code the logic which was previously implemented in QML.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -22,6 +22,7 @@ DEFINES += QT_STATICPLUGIN
 
 SOURCES += src/harbour-fernschreiber.cpp \
     src/appsettings.cpp \
+    src/chatpermissionfiltermodel.cpp \
     src/chatlistmodel.cpp \
     src/chatmodel.cpp \
     src/contactsmodel.cpp \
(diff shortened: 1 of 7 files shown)


jgibbon:
I haven't had the time to actually test this, yet. Will do after I can just grab a td build from the releases, because I'm a bit lazy at times. 

Preliminary "Review" without executing it:
- Looks really good! 
- Maybe the name "ChatForwardModel" isn't 100% accurate – since it's also used for prefilling the textarea after bot inline responses. Also it could be used (future) for sending text/files via the system's share thing should it ever be useable as a "receiver" in a harbour application.
Right now I can only think of pretty awkward names like "ChatPermissionFilterModel".

cheers!

monich:
`ChatPermissionFilterModel` sounds reasonable (albeit a bit long). I'll give a shot when I have time.

monich:
Renamed the class and rebased.

Wunderfitz:
I'd say let's merge it. Worked smoothly here on my device, thank you very much again! 

In case we have issues, we still can fix stuff later...

PR #469, opened 2021-12-12
Move Session options to Settings page

nephros:
As discussed briefly in the SFOS channel, moves the Session List and Logout button to the Settings.

People reported they didn't expect the button to be in the About page.  
While it's not really a "Setting", it's probably better located there.

RFC/Draft/PoC, ~~this first version has a problem with the ResposiveGrid and Landscape mode where the session list disappears...~~


qml/components/settingsPage/SettingsSession.qml
```diff
@@ -0,0 +1,248 @@
...
+                        activeSessionsItem.loaded = false;
+                        tdLibWrapper.getActiveSessions();
+                    }
+                }
+            }
+
+            ResponsiveGrid {
+                bottomPadding: Theme.paddingMedium
+
+                Loader {
```
monich, line 63: It doesn't make sense to place a single item into a grid.

nephros: It's two, but still you're right. A Flow then? See 185c8f7

monich: If it's always a single column then a Column would be appropriate. The purpose of that grid thing is to save vertical space in landscape (or on a large screen) by re-arranging items in two columns.

nephros: Which would still make sense as there are two items in that Accordeon (current session info and list of devices). The current variant with the Flow works as you describe, but is inconsistent with how the other Accordeons do it.

I'll look at a rework with ResponsiveGrid.

nephros: That was easy, Flow to ResponsiveGrid again. Squashed.

Wunderfitz:
Just had a short glance at it - I have to say, the new place definitely fits better. Thanks! However, the profile information and the logout button should be merged with the user profile section - that doesn't belong to sessions. And I'd probably not highlight the current app entry. Maybe consider keeping the layout of the current version.

nephros:
> Just had a short glance at it - I have to say, the new place definitely fits better. Thanks! However, the profile information and the logout button should be merged with the user profile section - that doesn't belong to sessions. And I'd probably not highlight the current app entry. Maybe consider keeping the layout of the current version.

OK, done!

Wunderfitz:
Looks much better now, thanks!

Some minor things: 
- Doubling the profile picture doesn't make sense
- BackgroundItem behind the button creates weird effects when the area around the button is clicked
- Loader for the phone numer is probably too much if the rest isn't loaded dynamically

But I simply merge it and do the rest myself. ;)

PR #470, opened 2021-12-18
Make "In reply to" properly react to taps

monich:
It was especially annoying that it wasn't reacting to long tap. You had to aim somewhere outside of it in order to open the context menu.


Wunderfitz:
Good catch, thank you very much!

PR #471, opened 2021-12-18
Remove (some) unused entries from QVariantMaps

monich:
This may save megabytes of heap depending on how many stickers you have (I have a lot 😀).


Wunderfitz:
Yep, I also noticed the new 'outline' attribute (wasn't there with TDLib 1.7.0), but of course I didn't think of memory consumption... Thanks for the automated removal!

PR #477, opened 2021-12-26
Fixed message reference in Replies chat

monich:
According to `td_api.tl`:

**`@reply_in_chat_id`** If non-zero, the identifier of the chat to which the replied message belongs; Currently, only messages in the Replies chat can have different reply_in_chat_id and chat_id


Wunderfitz:
Interesting, thanks!

PR #478, opened 2021-12-26
Don't try to load message with zero id

monich:
`chatPage.messageIdToShow` can be defined and yet be zero


Wunderfitz:
Good to know, thanks!

PR #482, opened 2022-02-20
Support textEntityTypeStrikethrough

monich:
It's not currently working because of [QtQuick bug](https://bugreports.qt.io/browse/QTBUG-72376) but maybe someday it will.


Wunderfitz:
Thank you very much! :)

PR #484, opened 2022-03-07
Update French translation

pherjung:
I didn't realized some sentence are related to the user. Here are the fix with a minor misspelling. Thank you for your work!


Wunderfitz:
thx as well!

PR #485, opened 2022-03-27
Adapt handling of sponsored messages to the new format

monich:
```
[D] TDLibReceiver::receiverLoop:185 - [TDLibReceiver] Raw result: {
    "@extra": "viewSponsoredMessage",
    "@type": "error",
    "code": 400,
    "message": "Failed to parse JSON object as TDLib request: Unknown class \"viewSponsoredMessage\""
}
```


Wunderfitz:
Mhmm, seems that the test channel no longer returns a sponsored message... Anyway, you seem to be right - the function has disappeared from [the official list](https://core.telegram.org/tdlib/docs/classtd_1_1td__api_1_1_function.html). Thanks!

PR #493, opened 2022-05-01
Made sure that context menu stays inside the screen area

monich:
When a list view is placed into another flickable, the context menu gets confused.


Wunderfitz:
Details matter, thank you very much! :)

PR #507, opened 2022-08-12
Improve message when search yields no results

nephros:
Currently, if one searches for contacts in the "New Chat" page, and no contacts match you get the message that  "You don't have any contacts".

This shows "No contacts found" if a search term has been entered.


Wunderfitz:
Makes perfect sense, sorry for the late reply...

PR #511, opened 2022-11-28
Always show user or group id on the info page

monich:
And copy it to the clipboard on tap. That's a useful piece of information other Telegram clients often prefer to hide.


Wunderfitz:
Sure, why not... Thanks and sorry for the late response.

Issue #512, opened 2022-12-09
Improve "chat has unread messages" in conversation overview


nephros:
When the conversation overview has many entries, the little "unread messages" bubble on the avatar become a little hard to spot.

This could be improved, e.g by using a different font color or background for such conversations.

Wunderfitz:
As the PR has been merged, we can possibly close this. ;)

PR #513, opened 2022-12-09
Highlight unread Converstations

nephros:
A proposed solution for Issue #512


monich:
IMO badge is a good enough mark. More fonts and colors don't necessarily make UI better.

On the other hand, since there's at least one person in the world who likes it this way, then if this behavior is configurable (and off by default) - why not.

Wunderfitz (changes requested):
I agree with @monich - I'm actually pretty happy with the way how it looks like at the moment. But if you make it configurable and not the default, I'm happy to merge it. (and sorry for the late response)

carlosgonz0:
Hi any ETA to merge this then new realease.? 
Thanks of advance all.

carlosgonz0:
@nephros where are you to push this.

Wunderfitz:
A little late, but as we do a new version, let's merge it. Thank you very much!

PR #514, opened 2022-12-16
Always append last message content to notifications

dscheinah:
I usually read and dismiss notifications on my smartwatch and only use the phone if I need to reply. This currently does not work for subsequent messages as the content is hidden.

With this PR the message content will always be appended to the notification. This allows me read all messages on the watch without using the phone if not needed.

I also removed the duplicated `nemoNotification->setBody(notificationBody);` line.

### DIFF
--- a/qml/components/settingsPage/SettingsBehavior.qml
+++ b/qml/components/settingsPage/SettingsBehavior.qml
@@ -165,6 +165,17 @@ AccordionItem {
                 }
                 Behavior on height { SmoothedAnimation { duration: 200 } }
             }
+
+            TextSwitch {
+                width: parent.columnWidth
+                checked: appSettings.notificationAlwaysShowPreview
+                text: qsTr("Always append message preview to notifications")
+                description: qsTr("In addition to showing the number of unread messages, the latest message will also be appended to notifications.")
+                automaticCheck: false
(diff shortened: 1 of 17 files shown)


Wunderfitz (changes requested):
I'm open to that, but only if
- it's made configurable
- it's not the default
Could you maybe include that in the settings? We already have a nice set of notification settings under 'Behavior'.
(And sorry for the late response...)

dscheinah:
No problem. I will see if I can do it in an acceptable way. But it may take a while because I'm quite busy at the moment.

How should I handle translations and do I need to take care of something special regarding persistence and migration?

dscheinah:
Alright, it's now added as a setting with off as default. Translations are mostly unfinished.

dscheinah:
Ok, fixed, so it should be ready now

jgibbon:
Thanks for your work – that seems like a worthwhile addition. Hadn't had the time to review yet, just wanted to show some appreciation.

Wunderfitz:
Far too late, but thanks! Let's merge and test. :)

dscheinah:
Great - thanks :) For testing I just realized that I obviously used my own build for a while now - because I already have the feature enabled.

PR #515, opened 2023-02-05
update en/de translations after latest PRs

jgibbon:


jgibbon:
oh you already did the translations… nevermind then :)

PR #516, opened 2023-02-06
99+ unread balloons

jgibbon:
A really small change:

- Unread formatting pushed to JS-Function
- Shows actual value until 999 (keeping the three char limit), then switches to 1k, 1k+, 2k, 2k+, ..., 9k+

Visually, it can be a tiny bit more crowded for some values, but I think it is nice to know the difference between, let's say, 150 or 5k. That could be the difference between "ok, I've got some time, let's read" and "oh, no way, mark all as read."


jgibbon:
To emulate higher numbers for review, just put a value = value * x; in the top of the function.

Wunderfitz:
Really late, but still: Thank you very much!

PR #521, opened 2023-10-04
Add option to suppress notification previews

nephros:
Currently untested, this was suggested in the German SFOS Telegram group.
Does not show message content in notifications.

A privacy/anti-embarrassment feature.

### DIFF
--- a/qml/components/settingsPage/SettingsBehavior.qml
+++ b/qml/components/settingsPage/SettingsBehavior.qml
@@ -137,8 +137,11 @@ AccordionItem {
 
             TextSwitch {
                 width: parent.columnWidth
-                checked: appSettings.notificationSuppressContent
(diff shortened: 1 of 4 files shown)


qml/components/settingsPage/SettingsBehavior.qml
```diff
@@ -135,6 +135,17 @@ AccordionItem {
                 }
             }
 
+            TextSwitch {
```
monich, line 138: Shouldn't it also be hidden if `notificationFeedback` is set to `NotificationFeedbackNone`, like the switch below?

carlosgonz0:
@nephros @monich  do you have a master of RPM to testing all this changes? Thanks of advance

Wunderfitz:
Once again quite late from my side, sorry and thanks!

PR #523, opened 2023-11-07
Move macros to yaml

nephros:
Define Macros in the .yaml file.

Due to a bug in spectacle, custom definitions in the .spec will be deleted on the next `specify` run.

See https://github.com/sailfishos/spectacle/issues/1 and https://forum.sailfishos.org/t/sdk-spectacle-deletes-custom-macros-in-rpm-spec/10868

This is not the case if macros are defined in the .yaml


Wunderfitz:
Thank you very much!

PR #526, opened 2023-11-19
Tweaked notification feedback settings UI

monich:
Particularly fixed the behavior of appearing/disappearing switches when notification feedback is being enabled/disabled. Organized feedback options into a column, slightly reduced the number of animations (animate the entire column rather than individual switches).


Wunderfitz:
thx

PR #527, opened 2023-11-19
Added UI for configuring session inactivity timeout

monich:
And a spinner while session information is being queried


Wunderfitz:
Nice enhancement, thx!

PR #530, opened 2023-11-21
Added "unread mention" indicator to the chat list

monich:
It's displayed in place of the "unread reaction" indicator. In case if there are both unread mentions and reactions, "unread mention" takes precedence. The end result is supposed to look like this:

image


Wunderfitz:
Great enhancement, thanks

PR #533, opened 2023-11-24
Ignore muted chats in unread count

nephros:
Luckily tdlib(warpper) already does deliver separate values.

This adds a setting, and when that's enabled, uses the alternate count.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -84,7 +84,7 @@ jobs:
           assets+=("-a" "$asset")
         done
         tag_name="${GITHUB_REF##*/}"
-        hub release create "${assets[@]}" -m "$tag_name" "$tag_name"
+        gh release create "$tag_name" "${assets[@]}" 
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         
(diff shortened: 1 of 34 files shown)


Wunderfitz:
Does Telegram restrict API keys nowadays? Or is it just that you haven't registered your own app yet?

Anyway, nice extension, but this only works if the TDLib database is used. In online-only mode this would fail, you also need to extend https://github.com/Wunderfitz/harbour-fernschreiber/blob/master/src/chatlistmodel.cpp#L573-L591

monich:
I wonder if Fernschreibers with different API keys can use the same tdlib db. That smells like potential trouble

monich:
My expectation is that "muted" means muted. In other words, silent. Not causing any notifications, sounds or other sorts of feedback. But not completely ignored.

I also expect that "unread" means unread. The ones that I haven't read. Not the ones that I muted.

FWIW Linux desktop client includes all unread messages into the overall count.

Wunderfitz:
I fully agree with @monich here, the concepts of "un/muted" and "un/read" are not interdependent IMHO. But I'm open to a settings switch (if the old behavior remains default) and the online-only mode is taken care of.

PR #534, opened 2023-11-26
[Trivial]: Don't own bindir

nephros:
This may seem like nitpicking, but the package should not own /usr/bin

### DIFF
--- a/rpm/harbour-fernschreiber.spec
+++ b/rpm/harbour-fernschreiber.spec
@@ -69,7 +69,7 @@ desktop-file-install --delete-original       \
 
 %files
 %defattr(-,root,root,-)
-%{_bindir}
+%{_bindir}/*
 %{_datadir}/%{name}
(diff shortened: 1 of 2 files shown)


rpm/harbour-fernschreiber.spec
```diff
@@ -69,7 +69,7 @@ desktop-file-install --delete-original       \
 
 %files
 %defattr(-,root,root,-)
-%{_bindir}
+%{_bindir}/*
```
Wunderfitz, line 72:
```suggestion
%{_bindir}/%{name}
```

I have to admit that I've never checked that, but wouldn't that be better? My knowledge doesn't go beyond some basic stuff from the manuals, so a second opinion would be appreciated...

Wunderfitz (commented):
I love nitpicking, but maybe we can improve it further... ;)

Wunderfitz:
Thx

Issue #502, opened 2022-06-05
Jump to post from quote


ApostolosB:
When you tap on a quoted message it opens in a "popup". It would be nice to have the ability to jump to that post from that "popup"

PR #538, opened 2023-11-28
Jump to post from quote

mbarashkov:
This PR fixes https://github.com/Wunderfitz/harbour-fernschreiber/issues/502 and jumps to the quoted message (instead of opening it in a popup (overlay) view).
If needed, the message is first loaded, and only then scrolled to.

### DIFF
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -468,8 +468,7 @@ ListItem {
                                         page.toggleMessageSelection(myMessage)
                                     } else {
                                         messageOptionsDrawer.open = false
-                                        messageOverlayLoader.overlayMessage = messageInReplyToRow.inReplyToMessage
-                                        messageOverlayLoader.active = true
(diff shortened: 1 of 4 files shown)


Wunderfitz:
Thx, generally open towards the change, but please don't remove it entirely. Please consider implementing it as an option and/or as a button "Jump to message" on the overlay.

mbarashkov:
@Wunderfitz I've added a setting for this (default is false to keep the current overlay behavior).

monich:
this doesn't always work for me 😐 sometimes it does, sometimes doesn't. there's nothing obvious in the log when it doesn't. overlay always works

mbarashkov:
@monich most likely it means the quoted message is already on screen, and there's no need to scroll to it. Please test carefully.

monich:
hmm, could be. then the target bubble needs to be highlighted for a brief period of time or something. right now nothing happens which gives an impression that... well, that nothing happens

Wunderfitz:
Just tested it, highlighting the target message seems to be pretty simply. Will add that after merging. Thanks, seems to work well - very lean solution! :)

PR #539, opened 2023-11-30
Do not animate message item content height

monich:
This animation looks really unnecessary or even annoying after switching between portrait and landscape


Wunderfitz:
I have to say I like it that way. :/ Maybe make it configurable?

monich:
I can do that. How would you like to call that option? I would call it "Randomly move list items around the screen" 🙂

monich:
Actually, I initially thought that this animation was a mistake, a copy/paste artifact or something. But if it's done on purpose then so be it. It's easier for me to just comment out a single line of QML on my device then to make this behavior configurable.

PR #540, opened 2023-12-01
Improve chat UI on tablets.

mbarashkov:
Improve chat UI on tablets in landscape mode by making messages narrower and limiting content items width as well.
Before this change, Tablet UI showed huge full-screen messages with attachments that very often took more than a screen height in "collapsed" state.

### DIFF
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -540,3 +540,7 @@ function getMessagesNeededForwardPermissions(messages) {
     }
     return neededPermissions
 }
+
+function isTablet(appWindow) {
+    return (appWindow.deviceOrientation & Silica.Orientation.LandscapeMask) || Silica.Screen.sizeCategory === Silica.Screen.Large || Silica.Screen.sizeCategory === Silica.Screen.ExtraLarge
+}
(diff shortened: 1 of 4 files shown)


qml/js/functions.js
```diff
@@ -540,3 +540,7 @@ function getMessagesNeededForwardPermissions(messages) {
     }
     return neededPermissions
 }
+
+function isTablet(appWindow) {
```
monich, line 544: `isTablet()` sounds misleading, I'd suggest `isWideScreen()` instead

mbarashkov: I suggest isLargeLandscapeScreen @monich

Wunderfitz: I'd also go for isWideScreen, but not really crucial for merging... ;)

Wunderfitz:
I like it, thank you very much!

mbarashkov:
@monich Actually it is not, I will make a follow-up improvement here.

Wunderfitz:
Funny, I thought it's intentional as it gives you some nice room for scrolling without the fear to hit this nasty scroll-to-top/bottom thingy. But if you want to work on it, nice! :)

monich:
I think a few more fixes are needed, see #546

PR #541, opened 2023-12-02
Fix broken signal-slot connection

monich:
```
QObject::connect: No such signal TDLibWrapper::chatAvailableReactionsUpdated(qlonglong, QString) in src/chatlistmodel.cpp:410
```


Wunderfitz:
Thx, it was quite late...

PR #543, opened 2023-12-03
Tweaked the logic of moving reactions into the view

monich:
There's no need to reposition list items if reactions bar are already fully visible.


Wunderfitz:
Details matter, thank you very much!

PR #548, opened 2023-12-04
Limit message width in widescreen UI.

mbarashkov:
This was intended to be included in https://github.com/Wunderfitz/harbour-fernschreiber/pull/540, but this change didn't make it into PR.

It further improved chat UI on tablets (or wide screens as we call them now) by limiting message width to 75% of the screen.

### DIFF
--- a/qml/components/MessageListViewItem.qml
+++ b/qml/components/MessageListViewItem.qml
@@ -130,6 +130,12 @@ ListItem {
     }
 
     function getContentWidthMultiplier() {
+        var type = myMessage.content["@type"];
+        if(type === "messagePoll") {
+            // We do not want to limit messagePoll content width on wide screens, it will not result in huge message items
+            // as other content types like image and video do.
+            return 1.0
+        }
         return Functions.isWidescreen(appWindow) ? 0.4 : 1.0
     }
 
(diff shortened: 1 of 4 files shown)


Wunderfitz:
It actually made it into the PR, but I reverted that after the discussion in https://github.com/Wunderfitz/harbour-fernschreiber/pull/540#issuecomment-1837311941 I guess, we need to come to a common conclusion here and I thought we agreed to restore the old behavior...

Wunderfitz (changes requested):
See my comment and the one in https://github.com/Wunderfitz/harbour-fernschreiber/pull/540#issuecomment-1837311941

mbarashkov:
Understood @Wunderfitz 
I will fix message aligning on tablets and update

mbarashkov:
@Wunderfitz 
PR updated, as per requested changes and tweaks.

monich:
Should this fix polls (which got squeezed too)

mbarashkov:
Will add poll fix as well

mbarashkov:
PR updated; polls now use full width of message.

Wunderfitz:
Did some additional tests with the current version and the Jolla 2 in widescreen. I have to admit that I like it as is, media sizes are limited and things have become even better with album support as it seems. In addition, I don't want to limit the message width in widescreen. Sorry, but closing for now.

Issue #549, opened 2023-12-04
Copy to clipboard doesn't work in landscape


mbarashkov:
Copy to clipboard does nothing if you select it from the additional landscape mode menu (from the More Options... panel).

Wunderfitz:
Currently can't reproduce it. On my tablet, the option is always in the first menu... Any hints how to reproduce it?

mbarashkov:
This is always reproduced on my own messages. Copy is only the 5th option.
You can move it to the extended menu manually and see for yourself that it's broken (see the pull request).

jgibbon:
Fix confirmed, thanks again!

PR #550, opened 2023-12-04
Fix Copy message to clipboard not working from extended menu

mbarashkov:
Fix issue https://github.com/Wunderfitz/harbour-fernschreiber/issues/549


jgibbon (approved):
I can confirm that this fix is working. Thank you!

PR #551, opened 2023-12-04
Show up to 6 items in popup menu of a message on large wide screen

mbarashkov:
On a large wide screen (tablet mode as I call it), we can easily show more popup menu items for a chat message context menu. I set this number to 6 - it looks nice on my tablet.


monich:
Yes, 6 is definitely way too much for a 21:9 screen

image

The limit could be based on the list view height divided by e.g. Theme.itemSizeSmall or whatever the menu item height is (keeping in mind that the list item itself should remain visible).

mbarashkov:
You're right, this has to be improved.
I'm using a 16:9 to inch tablet.

PR #553, opened 2023-12-07
Make the reaction button configurable

monich:


Wunderfitz:
Again it took a lot of time, thanks @monich

PR #556, opened 2024-01-01
Fix not all possible destinations shown to forward a message.

mbarashkov:
This simple PR lets you select from all eligible chats when forwarding a message.

### DIFF
--- a/qml/js/functions.js
+++ b/qml/js/functions.js
@@ -517,7 +517,7 @@ function handleErrorMessage(code, message) {
 }
 
 function getMessagesNeededForwardPermissions(messages) {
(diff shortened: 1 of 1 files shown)


monich:
Yeah, it was renamed by https://github.com/tdlib/td/commit/12d04ec684 somewhere between 1.8.11 and 1.8.12. Then comment just says "Rename can_send_messages to can_send_basic_messages" without explaining why. Just because they can :neutral_face:

monich (approved):
Works for me.

jgibbon:
Thanks @mbarashkov and @monich!

PR #563, opened 2024-01-31
Emoji update

### LINKED ISSUE #561: Twemoji update to Unicode 15.0
As we are aware, Twemoji became a bit out of date with the downfall of Twitter.

Community member dumol suggested using openmoji as an icon replacement: https://openmoji.org/library/

Steps to take:

1. Design/Taste check: Is openmojis design "nice enough" in general/are there any show-stoppers (contrast/background)? @Wunderfitz 
2. If 1. is OK: Adaptation of our qml integration @jgibbon 

An evaluation of Determinants c++ emoji parser (also twemoji) would be nice, but I guess we won't have time for that in the near future. https://github.com/R1tschY/harbour-determinant/blob/63204c9769084a5bf3342a74e5ea4d039c3938df/src/emojiparser.cpp

cheers

jgibbon:
- adds newer emoji svg files from twemoji fork https://github.com/jdecked/twemoji
- updates twemoji.js regex
- updates emojis.db to one generated by https://github.com/jgibbon/fernie-emojidb (dropping the .svg in every db entry to save a few KB) matching all available icons with https://unicode.org/Public/emoji/15.1/emoji-test.txt

fixes #561


jgibbon:
As a small help for next updates (15.1 is out!)

The "reference regex" can be found inside twemoji-parser: https://github.com/jdecked/twemoji-parser/blob/main/src/lib/regex.js

Issue #566, opened 2024-05-14
Album feature has broken "scroll to quoted message" 


jgibbon:
display indexes of messages are now shifted by number of album-embedded images

(text copy/pasted from telegram group bug report)

PR #567, opened 2024-05-14
Fix scroll to quoted message

mbarashkov:
This pull request fixes issue #566 by calculating the proper index of message that we want to go to, substracting number of additional album-contained messages from it.


mbarashkov:
Improving a bit and resubmitting shortly

PR #568, opened 2024-05-14
Fix scroll to quoted message

mbarashkov:
This pull request fixes issue #566 by calculating the proper index of message that we want to go to, substracting number of additional album-contained messages from it.


jgibbon:
Hi @mbarashkov, thanks for the PR! 

It may be easier to just ask the proxy model for the id without any c++ changes, because that is the model for chatView:

``` var index = chatModel.getMessageIndex(chatPage.messageIdToScrollTo);
            var proxyIndex = chatProxyModel.mapRowFromSource(index, -1);
            if(index !== -1) {
                chatPage.messageIdToScrollTo = "";
                chatView.scrollToIndex(proxyIndex);
                navigatedTo(proxyIndex);
            }
 ```

Or did you try that and it didn't work?

cheers

mbarashkov:
@jgibbon 
I just tried it, and mapRowFromSource doesn't seem to work correctly in a very simple test.
Actually this method just infinitely loops.

jgibbon:
Hm, weird. 
If I'm not mistaken that should be exactly what it's intended for – in that case there could be another bug… I'll have another look at this.

Wunderfitz:
My tests are indicating that @jgibbon's solution actually works. I'll commit this for now and close this. In case future tests reveal that there are infinite loops, I'll dig this one out again. In any case, thanks for your efforts, @mbarashkov and @jgibbon.

Wunderfitz:
See https://github.com/Wunderfitz/harbour-fernschreiber/commit/cb09be5fd4bddc93f55f323e8809369f553c987b

PR #569, opened 2024-05-15
Force xz rpm compression

monich:
That improves the chances of rpm being installable on a system that doesn't support zstd compression (like my tablet)

It's not clear to me why to maintain both yaml and spec files but if the owner likes it this way then so be it


jgibbon:
Looks good to me. But https://forum.sailfishos.org/t/sfos-4-5-0-switches-rpm-compression-what-does-that-mean-for-existing-rpms/13153/18 argues that w9 may be the way to go, what's your opinion on that?

cheers

monich:
I have no preference wrt compression ratio, I picked w6 because it was the default before the switch

jgibbon:
Me neither, so I think w6 is okay :)

PR #570, opened 2024-06-15
Avoid UI updates in Chat selection screen when forwarding a chat message

mbarashkov:
This pull request removes the annoying and useless update of the list of chats when you select a chat to forward a message to (these constant updates actually stop you from easily selecting a chat to forward your message to).
It creates a "frozen" copy of chatlistmodel to be used by the ChatPermissionFilterModel (which is only used by the "select a chat..." screen so far).


Wunderfitz:
I'm not the biggest fan of cloning data for such a purpose and it's also hard to test for me as my groups don't have that many messages. I park this for now and do some cosmetic changes to the chat list model (e.g. animation when a order change occurs).

Wunderfitz:
See 35978b5107905d9658223276a480d037917db5a8

PR #571, opened 2024-06-15
tdlibwrapper refactoring and optimization

mbarashkov:
A few small changes to tdlibwrapper:
1. Refactored misleadingly names variables allUsers and  to more proper names usersById and usersByName.
2. Fixed handleUserUpdated function as it mixed up userInformation parameter with userInformation member variable.
3. Made a dedicated method updateUserInformation to update both usersById and usersByName
4. Eliminated unneeded status updates when user status didn't really change (those tend to happen a lot).

### DIFF
--- a/src/tdlibwrapper.cpp
+++ b/src/tdlibwrapper.cpp
@@ -1573,17 +1573,17 @@ QVariantMap TDLibWrapper::getUserInformation()
 QVariantMap TDLibWrapper::getUserInformation(const QString &userId)
 {
     // LOG("Returning user information for ID" << userId);
-    return this->allUsers.value(userId).toMap();
+    return this->usersById.value(userId).toMap();
 }
 
 bool TDLibWrapper::hasUserInformation(const QString &userId)
 {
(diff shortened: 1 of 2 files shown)


monich:
As a side note, I find it weird that `QString`'s and `QVariant`'s are being used to store the user id (which is `int53` according to [td_api.tl](https://github.com/tdlib/td/blob/master/td/generate/scheme/td_api.tl)), `qlonglong` would be far more efficient both in terms of memory usage and overall efficiency (e.g. hashing and comparison). The change itself looks ok to me.

mbarashkov:
@monich it doesn't make much sense, however, those go quite deep in the code, so I don't want to fix it right now, there are more obvious areas to improve.

monich:
yes, I agree

monich (approved):
Let's merge it then

monich:
Egh, the commit message wasn't formatted in a sane way... Too late to fix this one but let's pay more attention to that in the future

PR #572, opened 2024-06-21
Save power by adding a mandatory sleep between tdlib requests

mbarashkov:
0.02 second in foreground more and 0.1 second in background mode.
This also considerably improves UI smoothness in my tests.

### DIFF
--- a/qml/harbour-fernschreiber.qml
+++ b/qml/harbour-fernschreiber.qml
@@ -30,14 +30,6 @@ ApplicationWindow
     cover: Qt.resolvedUrl("pages/CoverPage.qml")
     allowedOrientations: defaultAllowedOrientations
 
-    onApplicationActiveChanged: {
-        if(applicationActive) {
-            tdLibWrapper.handleForeground()
-        } else {
-            tdLibWrapper.handleBackground()
-        }
-    }
-
     Connections {
         target: dBusAdaptor
         onPleaseOpenMessage: {
--- a/qml/harbour-fernschreiber.qml
+++ b/qml/harbour-fernschreiber.qml
@@ -30,6 +30,14 @@ ApplicationWindow
     cover: Qt.resolvedUrl("pages/CoverPage.qml")
     allowedOrientations: defaultAllowedOrientations
 
+    onApplicationActiveChanged: {
+        if(applicationActive) {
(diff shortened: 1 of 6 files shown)


src/tdlibreceiver.cpp
```diff
@@ -205,6 +214,7 @@ void TDLibReceiver::receiverLoop()
           VERBOSE("Raw result:" << receivedJsonDocument.toJson(QJsonDocument::Indented).constData());
           processReceivedDocument(receivedJsonDocument);
       }
+      sleep(this->powerSavingMode ? POWERSAVING_TDLIB_REQUEST_INTERVAL : NORMAL_TDLIB_REQUEST_INTERVAL);
```
monich: What is this sleep that takes `double` as an argument? I suggest to use [msleep](https://doc.qt.io/qt-5/qthread.html#msleep) instead, at least it's clear what it's doing and which units it takes.

qml/harbour-fernschreiber.qml
```diff
@@ -30,6 +30,14 @@ ApplicationWindow
...
     allowedOrientations: defaultAllowedOrientations
 
+    onApplicationActiveChanged: {
+        if(applicationActive) {
+            tdLibWrapper.handleForeground()
+        } else {
+            tdLibWrapper.handleBackground()
+        }
+    }
+
```
monich: There's no need to involve QML in this, the native `TDLibWrapper` object can directly receive `QGuiApplication::applicationStateChanged` signals from `qGuiApp`.

mbarashkov:
Will address @monich

mbarashkov:
@monich feedback addressed

mbarashkov:
Also my tests showed that throttling requests in foreground is bad because it affects app performance significantly. So we should do it in background.

monich (approved):
LGTM

monich:
Oh, and one more thing, Could you please squash the commits and format the commit message so that it looks nice in the git history.

mbarashkov:
@monich while fighting with the Github Desktop app etc I've completely messed up the branch. I will make a new pull request.

mbarashkov:
@monich new PR is https://github.com/Wunderfitz/harbour-fernschreiber/pull/573

PR #573, opened 2024-06-30
Save power by adding delay between network requests in background.

mbarashkov:
This is a remake of https://github.com/Wunderfitz/harbour-fernschreiber/pull/572

### DIFF
--- a/qml/pages/OverviewPage.qml
+++ b/qml/pages/OverviewPage.qml
@@ -88,7 +88,8 @@ Page {
     }
     Timer {
         id: updateSecondaryContentTimer
-        interval: 600
+        interval: 1500
+        repeat: false
         onTriggered: {
             chatListModel.calculateUnreadState();
             tdLibWrapper.getRecentStickers();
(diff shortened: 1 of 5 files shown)


qml/pages/OverviewPage.qml
```diff
@@ -88,7 +88,8 @@ Page {
     }
     Timer {
         id: updateSecondaryContentTimer
-        interval: 600
+        interval: 1500
+        repeat: false
```
monich, line 92: There's no need for this, `repeat` defaults to `false`

mbarashkov: > There's no need for this, `repeat` defaults to `false`

Above this another method states "repeat: false" expicitly. I assumed that's a code style. After all it's better readable this way..

monich: And is this necessary at all, to increase the interval? This is a one-time action, it won't have any impact on power usage or anything like that.

mbarashkov: Well, yes, it's logically unrelated, I've just noticed it improves app startup performance.. Ok, I will redo the PR again.

mbarashkov:
@monich change removed in https://github.com/Wunderfitz/harbour-fernschreiber/pull/574

PR #574, opened 2024-07-02
Save power by adding delay between network requests in background.

mbarashkov:
This is a remake of https://github.com/Wunderfitz/harbour-fernschreiber/pull/573


Wunderfitz:
Thank you very much - again with a lot of delay. As @monich already approved #572 and this one doesn't include the interval that was discussed, merging this one.

PR #576, opened 2024-12-28
Bugfixes albums; notches

jgibbon:
Fixes downloads in message media albums views, adds rudimentary notch support for jolla c2


qml/components/messageContent/mediaAlbumPage/VideoComponent.qml
```diff
@@ -32,7 +32,7 @@ Video {
 
         width: parent.width //don't use anchors here for easier custom scaling
         height: parent.height
-//        highlighted: parent.highlighted
+        highlighted: parent.highlighted || false
```
monich: hmm, the `|| false` part seems redundant to me. what's the point of that?

jgibbon: I think there was some warning I wanted gone at some point, I'll have to check that. Thanks!

jgibbon:
I somehow messed up the top chat page padding conditional for devices without notch in the initial implementation. Should be fixed now.

qml/components/messageContent/mediaAlbumPage/VideoComponent.qml
jgibbon: Video doesn't have the highlighted property anyway, so this might have started as a copy/paste error. I have removed the line.

qml/components/messageContent/MessageContentFileInfoBase.qml
```diff
@@ -57,7 +57,7 @@ MessageContentBase {
         width: parent.width
         height: Theme.itemSizeLarge
         Loader {
-            active: contentItem.thumbnail || contentItem.minithumbnail
+            active: (contentItem.thumbnail || contentItem.minithumbnail) || false
```
monich, line 60: another `|| false` :)

jgibbon: Thanks for checking further! 
In this case both thumbnail and minithumbnail could possibly be undefined. 
I think this one could stay, possibly even with a `!!` before the `(` to explicitly "cast" to bool… what do you think?

monich: Agh, you're right -  in QML  (and JavaScript in general) the `&&` and `||` operators return the value of one of the operands rather than a boolean (which is a bit counter-intuitive, I keep forgetting about it when reading JavaScript code). Then it's fine the way it is. `!!` would work too I guess, it's up to you - whatever you like better.

jgibbon: I'll just keep it, I think – this works and the change wouldn't exactly improve readability. 
Thank you for taking the time to recheck!

Wunderfitz:
Again took a while, but merging it finally. Thanks!

PR #577, opened 2025-01-05
Show duration and success of (video)calls

simonschmeisser:
I figured it would be nice to see in history if I missed a call or actually talked for some while. Not super important but a simple tiny improvement to get started


simonschmeisser:
Should I include updated translation files or is that handled by CI?

PR #579, opened 2025-06-07
Hungarian Translations

Lundrin:
Completed the missing lines of the hungarian translations


Wunderfitz:
I'm very late, but still: Thank you very much!

PR #581, opened 2026-08-09
Add admin member actions to the group members list

JimKnopfIoT:
Small, self-contained feature: managing group members from inside Fernschreiber
instead of having to reach for another client. Five source files plus the
translations, no new dependencies, no build-system changes beyond one
`DISTFILES` line.

This is deliberately kept separate from the calls PR — the two have nothing to do
with each other, and this one should not have to wait for that discussion.

## What it does

Long-pressing a member in a group's member list opens a context menu. Which
entries appear depends on what *you* are allowed to do and on what the target
is — admins and the creator are never offered, and nothing shows up at all
unless you have the right to restrict members.

| Entry | Effect |
|---|---|
| **Member Permissions** | opens a dialog with 13 granular permission switches |
| **Revoke Write Permission** | quick restrict, without opening the dialog |
| **Remove Restrictions** | back to the chat's default permissions |
| **Remove from Group** | removes the member (remorse timer) |
| **Remove and Delete All Messages** | removes and revokes their messages (remorse timer) |

The permissions dialog is prefilled with the member's *current* rights and each
switch is **capped by the chat's default permissions**: a right the group does
not grant at all is shown disabled, with the reason spelled out
("Not allowed by the group's default permissions"), rather than silently doing
nothing when toggled.

## How it works

Pure TDLib signalling — `setChatMemberStatus` and `banChatMember` on
`TDLibWrapper`, plus the QML. No new libraries, no C++ beyond those two wrappers.

One thing worth knowing for review: **`updateChatMember` is only delivered to
bots**, so a user client never gets told that its own change went through. The
member list is therefore updated optimistically in the model after the request is
sent. If you would prefer an explicit re-fetch of the member instead, that is an
easy change.

## A caveat that has since resolved itself

When this PR was opened, the `tdlib/include` headers in the repo were **older
than the shipped `libtdjson.so` 1.8.21**, so the runtime JSON did not match them:
admin rights arrived nested in `status.rights` rather than flat on `status`, and
`chatPermissions` already used the granular field names. I read the real names
with `strings` on the shipped `.so` rather than trusting the headers.

Since the 1.8.67 update on master, headers and library agree again —
`chatMemberStatusAdministrator` nests its rights in `rights`, and
`chatPermissions` carries sixteen granular `can_*` fields. The code still reads
both layouts and each permission keeps a `legacy:` fallback name, so a member
restricted by an older TDLib is shown correctly either way. What the rebase did
change is in the comment below.

## Translations

German and English are translated; the other languages carry the new strings as
`type="unfinished"`. The `.ts` diff is exactly what your own build produces —
I added the entries by hand first, then let a `mb2 build` run its `lupdate` and
committed that result, so the files are canonical and your working copy stays
clean after the first build. Every file gains the same 118 lines and loses none.

## Tested

On device (Jolla phone, SailfishOS 5.2.0.17 aarch64), after the 1.8.67 rebase:
all five actions against a real group, including the permission caps and both
remorse timers. That run found the two bugs described in the comment below;
both are fixed and were re-tested on the device afterwards.

Built from this branch's exact commit on top of the current master:
`SailfishOS-4.6.0.13-aarch64`, clean, with `ChatMemberPermissionsPage.qml` in the
RPM. The older three-target matrix (4.6.0.13 aarch64 + armv7hl, 5.0.0.62 aarch64)
was measured on the pre-rebase commit.

Happy to adjust the wording, the menu structure, or the optimistic-update
approach if you would rather have it differently.

### DIFF
--- a/harbour-fernschreiber.pro
+++ b/harbour-fernschreiber.pro
@@ -134,6 +134,7 @@ DISTFILES += qml/harbour-fernschreiber.qml \
     qml/js/functions.js \
     qml/pages/ChatInformationPage.qml \
     qml/pages/ChatPage.qml \
+    qml/pages/ChatMemberPermissionsPage.qml \
     qml/pages/ChatSelectionPage.qml \
     qml/pages/CoverPage.qml \
     qml/pages/DebugPage.qml \
--- a/qml/components/chatInformationPage/ChatInformationTabItemMembersGroups.qml
+++ b/qml/components/chatInformationPage/ChatInformationTabItemMembersGroups.qml
@@ -71,10 +71,101 @@ ChatInformationTabItemBase {
             text: (chatInformationPage.isPrivateChat || chatInformationPage.isSecretChat) ? qsTr("You don't have any groups in common with this user.") : ( chatInformationPage.isChannel ? qsTr("Channel members are anonymous.") : qsTr("This group is empty.") )
         }
         delegate: PhotoTextsListItem {
+            id: memberListItem
             pictureThumbnail {
                 photoData: user.profile_photo ? user.profile_photo.small : null
             }
             width: parent.width
+
+            // Admin actions on other members: only offered if we can restrict
+            // members and the target is a regular/restricted member (admins and
+            // the creator can't be managed this way). Newer TDLib nests the
+            // admin rights in status.rights, older TDLib has them flat.
+            readonly property bool ownUserCanRestrictMembers: (chatInformationPage.isSuperGroup || chatInformationPage.isBasicGroup)
+                                                              && chatInformationPage.groupInformation.status
+                                                              && ( chatInformationPage.groupInformation.status["@type"] === "chatMemberStatusCreator"
+                                                                  || ( chatInformationPage.groupInformation.status["@type"] === "chatMemberStatusAdministrator"
+                                                                      && ( chatInformationPage.groupInformation.status.rights
+                                                                          ? chatInformationPage.groupInformation.status.rights.can_restrict_members
+                                                                          : chatInformationPage.groupInformation.status.can_restrict_members ) ) )
+            readonly property bool memberIsManageable: member_id.user_id !== chatInformationPage.myUserId
+                                                       && ( model.status["@type"] === "chatMemberStatusMember"
+                                                           || model.status["@type"] === "chatMemberStatusRestricted" )
+
+            menu: (ownUserCanRestrictMembers && memberIsManageable) ? memberContextMenu : null
+
+            Component {
+                id: memberContextMenu
+                ContextMenu {
+                    MenuItem {
+                        visible: chatInformationPage.isSuperGroup
+                        text: qsTr("Member Permissions", "edit a group member's individual permissions")
+                        onClicked: {
+                            var setIndex = index;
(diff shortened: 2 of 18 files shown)


JimKnopfIoT:
Rebased onto master (0.19 / TDLib 1.8.67). Nothing conflicted, but the update had
broken the dialog in a way that does not announce itself:

* 1.8.67 renamed `can_add_web_page_previews` to `can_add_link_previews` and
  `can_manage_topics` to `can_create_topics`. `EditGroupChatPermissionsColumn.qml`
  moved with the rename, this page did not, so its switch wrote a field TDLib no
  longer knows.
* The heavier one: the dialog rebuilt `chatPermissions` from its own thirteen
  switches alone. A field left out of that object counts as `false`, so every save
  would have quietly revoked `can_react_to_messages`, `can_edit_tag` and
  `can_create_topics` - rights the dialog never showed and the admin never touched.
  It now carries over everything it has no switch for: from the member's current
  rights, or from the chat's default permissions for a member who is not
  restricted yet.

The older names are still read as a fallback, so a member restricted by an earlier
TDLib is still shown correctly.

Built clean against `SailfishOS-4.6.0.13-aarch64`. The on-device run in the
description predates this, so the two points above have not been through a device
test yet — I will report back once they have.

JimKnopfIoT:
The device test happened, and it turned up two real bugs. Both fixed, both
re-tested on the device afterwards.

Neither was in the request: TDLib answers every `setChatMemberStatus` with `Ok`
and the change is on the server. What failed both times was the optimistic
update that has to stand in for the missing `updateChatMember` — and in both
cases the cause was the same, the context menu tearing down the very delegate
the follow-up work was hanging off.

**1. The permission dialog looked like it did nothing** (`dd6ccd3`)

Save, reopen, and the old rights were back. Leaving the member list and
entering it again showed the new rights, because the list is then fetched
afresh.

```
ChatInformationTabItemMembersGroups.qml:120: TypeError:
Cannot read property 'membersList' of undefined
```

The delegate is gone by the time the dialog is accepted, so `pageContent` and
`index` no longer resolve inside the callback. Both are read before the dialog
is pushed now and captured by the closure.

**2. Removing a member did nothing at all** (`d0eeb08`)

The remorse countdown appeared and then stood still. Nothing was sent, the
member stayed in the group, and the row kept a frozen overlay where it had
been. `ListItem.remorseAction()` parents the countdown to the delegate's
`contentItem`:

```qml
function remorseAction(text, action, timeout) {
    return Remorse.itemAction(contentItem, text, action, timeout)
}
```

and the context menu is tearing that delegate down at the same moment, so the
countdown never reaches its callback. It is a `RemorsePopup` on the page now,
which outlives the row it is about. The callback resolves the wrapper, the ids,
the model and the index while the delegate is still alive, so nothing in it
depends on a context that is about to disappear.

The two one-tap actions were never affected: they update the model
synchronously inside `onClicked`.

Tested on a Jolla phone, SailfishOS 5.2.0.17 aarch64: permissions stand right
after saving, and a removed member disappears from the list as the remorse runs
out.

Wunderfitz:
Thank you very much. Made some tiny adjustments that surfaced during review. Will merge now. :)

PR #582, opened 2026-08-09
Add a setting to hide "Delete Chat" from the chat menu

JimKnopfIoT:
Tiny change, and it follows on from a conversation we had a while ago about how
easy it is to delete a chat by accident.

0.18 already improved this a lot — the confirmation dialog you added on top of
the remorse timer was not there in 0.17. This is about the remaining case: the
entry being *present* at all.

## The reasoning

Deleting a chat is irreversible, and unlike everything else in that pulley menu
it is not something most people ever do from the phone — it is a desktop task.
Yet it sits in the same short menu as Mute Chat and Search in Chat, which are
daily actions. A confirmation only helps someone who registers what they
triggered; someone who did not mean to open that flow at all tends to dismiss
whatever appears.

Rather than change the behaviour for everyone, this just lets people who never
delete chats from the phone take the entry out of their way.

## What it does

New `AppSettings` property `showDeleteChat`, **default `true`** — without
touching the setting, nothing changes for anyone. A switch in
**Settings → Behavior**, next to the other chat-behaviour options:

> **Show "Delete Chat" in the chat menu**
> Deleting a chat is irreversible and rarely needed on the go. Turn this off to
> keep the entry out of the pulley menu.

The menu entry keeps its confirmation dialog and remorse timer unchanged when
shown. Four files plus translations; the property follows the exact shape of the
twelve existing boolean settings.

## Built

From this branch's exact commit with the Sailfish SDK (4.6.0.13):
**aarch64 OK (166 s)** and **armv7hl OK**. New strings are in the source `.ts`,
`-en` and `-de`; the other languages carry them as untranslated, exactly as the
build's own `lupdate` generated them.

## If you would rather not

Entirely understandable if you do not want another switch in Settings — the
alternative would be moving the entry to the bottom of the pulley menu, so that
the shortest pull no longer lands on the most destructive action. Happy to
rework it that way, or to drop it.

### DIFF
--- a/qml/components/settingsPage/SettingsBehavior.qml
+++ b/qml/components/settingsPage/SettingsBehavior.qml
@@ -94,6 +94,17 @@ AccordionItem {
 
             TextSwitch {
                 width: parent.columnWidth
+                checked: appSettings.showDeleteChat
+                text: qsTr("Show \"Delete Chat\" in the chat menu")
+                description: qsTr("Deleting a chat is irreversible and rarely needed on the go. Turn this off to keep the entry out of the pulley menu.")
+                automaticCheck: false
+                onClicked: {
+                    appSettings.showDeleteChat = !checked
+                }
+            }
+
+            TextSwitch {
+                width: parent.columnWidth
                 checked: appSettings.notificationAlwaysShowPreview
                 text: qsTr("Always append message preview to notifications")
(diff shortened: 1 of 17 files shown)


Wunderfitz:
Thank you very much, of course merging it. :)

PR #583, opened 2026-08-09
Add Telegram voice & video calls (tgcalls + tg_owt)

JimKnopfIoT:
**In one sentence:** this adds working 1:1 voice *and* video calls, entirely
behind a build flag — without `CONFIG+=voicecalls` the app you ship today is
unchanged.

I know this is a large PR landing unannounced, so the most important part first:
**how you can ignore most of it.** Everything that needs a WebRTC stack lives
behind `voicecalls {}` in the `.pro`, in a separate `voip.pri`, and in
`src/voip/`. A stock `mb2 build` never sees it. I have verified that explicitly
(see the build matrix below) rather than just asserting it.

## Why this needs a second library at all

TDLib does call **signalling only** — `createCall` / `acceptCall` /
`discardCall`, the DH exchange, protocol negotiation, the relay list. It never
produces or consumes a single media frame. The media has to come from
**tgcalls**, which sits on Telegram's WebRTC fork (**tg_owt**). That is why
Fernschreiber never had calls, and it is the only part of this that is genuinely
heavy.

The signalling side is kept deliberately **media-library agnostic**:
`TDLibWrapper` / `TDLibReceiver` gain the call methods and updates and know
nothing about tgcalls. If you ever wanted a different media backend, that layer
would stay as it is.

## Status

| | |
|---|---|
| Scope | 1:1 voice and video. **No group calls / voice chats.** |
| Tested on | Jolla phone, Sailfish OS 5.2.0.17; earlier on an Xperia 10 III, 5.1.0.11 |
| Interop | official Telegram Android and Desktop, both call directions, voice and video |
| Media architecture | aarch64 (see the armv7hl note below) |
| Encryption | the usual Telegram E2EE call path; SAS emojis are shown in the UI |

## Build matrix

Built from this branch's exact commit (`295d796`), on top of the current master,
all four measured today on the same machine:

| build | target | result |
|---|---|---|
| `mb2 build` (stock) | 4.6.0.13 aarch64 | OK, 172 s — 0 call symbols in the binary |
| `mb2 build` (stock) | 4.6.0.13 armv7hl | OK, 159 s — same |
| `mb2 build` (stock) | 5.0.0.62 aarch64 | OK, 177 s — same |
| `mb2 build -- --with voicecalls …` | 4.6.0.13 aarch64 | OK, 342 s — 30437 tgcalls symbols, the policy file and 16 multimedia libraries packaged |

Two SDK generations apart, so nothing here is tied to one release. Checked on the
RPM itself with `rpm -qlp` and `rpm -qRp` rather than on the build's exit code:
none of the bundled libraries appears in the package's requires, so nothing is
demanded that no repository can satisfy. Only one file of this PR reaches a stock
build — `qml/components/CallOverlay.qml`, which travels with the `qml` directory
and is never loaded there, because `voiceCallsAvailable` is false. The two
`libtdjson` entries a stock build does carry are upstream's own.

Enabling the calls in an RPM build is wired up as a plain rpmbuild conditional:

```sh
mb2 -t <target> build -- --with voicecalls \
    --define "tg_owt_root /path/to/tg_owt" \
    --define "tgcalls_root /path/to/tgcalls" \
    --define "openh264_root /path/to/openh264"
```

It is declared as `QMakeOptions` in `rpm/harbour-fernschreiber.yaml`, so it
survives spec regeneration, and the `%files` entry for the policy rule is guarded
by the same conditional.

One packaging detail worth knowing: **the multimedia libraries tg_owt needs
travel inside the package**, next to `libtdjson` in the binary's rpath
directory, with `__requires_exclude` keeping them out of the RPM's
dependencies. That is `libopenh264.so.8` — Sailfish packages openh264 nowhere,
for patent reasons, while tg_owt hard-requires it — and with it `libvpx`,
ffmpeg, opus, ogg, vorbis, theora, webp, openjpeg and speex. Their sonames
differ between Sailfish releases, so a package built against one release would
not resolve on another: 5.2 ships `libvpx.so.12` where the 4.6 SDK links
`.so.9`. Without this the calls RPM would declare dependencies no repository
can satisfy and simply refuse to install.

An earlier version of this PR bundled only openh264 and took `libvpx` from the
system by soname. That is what the commit "Bundle the multimedia stack so one
RPM works across Sailfish releases" replaces.

The point of rows 3 and 4: **the PR does not touch your existing build on either
architecture.** The calls only compile in when explicitly asked for.

## What is in here

| Path | Role |
|---|---|
| `src/tdlibwrapper.*`, `src/tdlibreceiver.*` | TDLib call signalling. Media-library agnostic. |
| `src/voip/voipmanager.*` | tgcalls instance lifetime, descriptor (relays, key, protocol), signalling relay, QML-facing API |
| `src/voip/sailfishinterface.*` | tgcalls `PlatformInterface` for Sailfish: codec factories, capturer creation |
| `src/voip/callcameragrabber.*`, `videocapturer.*`, `videotracksource.h` | camera: QtMultimedia → I420 → WebRTC, incl. device orientation |
| `src/voip/videorenderer.*` | frames → QML `VideoOutput` |
| `src/voip/callaudiorouter.*` | moves the WebRTC playout to the call sink and unmutes it, via in-process libpulse |
| `qml/components/CallOverlay.qml` | in-call UI: peer, state, SAS emojis, remote video + local PiP, mute / hang up / flip |
| `voip.pri`, `doc/voicecalls.md` | build glue and build instructions |
| `pulse/harbour-fernschreiber.conf` | audio policy rule, see below |
| `translations/` | new strings in the source `.ts`, `-en` and `-de` |

`.desktop` gains exactly one Sailjail permission: `Camera`.

## Incoming calls

An incoming call rings through **ngfd's `voip_ringtone` event**, not through
Fernschreiber's notification settings. That distinction matters in practice: a
call is not a message, so someone who has turned message tones off — a perfectly
normal setup — must still hear the phone ring. Going through ngfd also means
volume, vibration and silent mode follow the system profile, exactly like the
built-in dialer. The ringtone is tied to the call state rather than to the
overlay being visible, so it also rings while the app sits in the background.

The app **deliberately does not raise itself to the foreground** for an incoming
call — you open it yourself to answer. Grabbing the screen unasked seemed like
the kind of behaviour that should be your call rather than mine. If you would
rather have it behave like the dialer, say so and I will add it — either raising
the window, or a high-priority notification with answer/decline actions as the
gentler middle ground.

## Two things I would like your decision on

**1. The PulseAudio policy rule.** Without it the remote party is inaudible, so
it is not optional — but it installs to `/etc/pulse/xpolicy.conf.d/`, outside
the app tree, which a Harbour build would flag.

The reason it is needed: tgcalls names its streams `WEBRTC VoiceEngine`, the
Sailfish policy sorts unknown streams into the `othermedia` group, and that group
carries `cork_stream` — so the moment the call's microphone capture starts, the
policy corks the incoming-audio playback stream. The rule moves the WebRTC
streams into the telephony `call` group, which is what a real call uses. The
`INSTALLS` entry is inside the `voicecalls {}` block, so a stock build is
unaffected. If you would rather have a separate configuration sub-package, say
the word and I will restructure it.

**2. tg_owt and tgcalls are not vendored.** They are expected out-of-tree via
`TG_OWT_ROOT` / `TGCALLS_ROOT` (defaulting to a sibling directory). Vendoring
them as submodules would make the build reproducible for everyone, but that is a
packaging decision that felt like yours to make rather than mine to slip in.
`voip.pri` fails fast with an actionable message if the trees are missing or
unbuilt, instead of dumping thousands of unresolved WebRTC symbols at link time.

## About armv7hl

The **app** builds on armv7hl, verified above. The **media stack** was only ever
built for aarch64, so calls on armv7hl are untested: it would need a tg_owt built
for that target. `voip.pri` therefore emits a warning rather than pretending it
works, and the aarch64-specific bits (camera rotation values, resolution pinning)
are commented with the device they were measured on.

## Trying it out

`doc/voicecalls.md` has the exact revisions of the two forks, the cmake
invocation for tg_owt, and the `mb2` line. The forks are the ones RooTelegram and
Yottagram use; they carry the Sailfish/Halium fixes upstream tgcalls lacks. The
audio-routing approach is adapted from RooTelegram (GPL-3.0) and attributed in
the source.

## Not included

- group calls / voice chats (the tgcalls group sources are deliberately excluded)
- call history UI beyond the `messageCall` service-message text, which now
  names how the call ended (taken from #577, see the comment below)
- hardware video encoding — the encoder is software VP8/VP9. Decoding accepts
  everything the builtin factory offers, including H264/H265, so official
  clients (which prefer H264) are received reliably.

Happy to split this up, rework any part of it, or drop pieces you do not want.

### DIFF
--- a/doc/voicecalls.md
+++ b/doc/voicecalls.md
@@ -1,0 +1,188 @@
+# Voice and video calls
+
+Fernschreiber can be built with 1:1 voice and video calls. The feature is
+**opt-in at build time** (`qmake CONFIG+=voicecalls`); without that flag nothing
+in this document applies and the build is byte-for-byte the one you know.
+
+## Why an extra library is needed
+
+TDLib does call **signalling only** — `createCall` / `acceptCall` /
+`discardCall`, the DH key exchange, protocol negotiation and the relay list. It
+never produces or consumes a single audio or video frame. The media has to come
+from a separate library, and for Telegram that is **tgcalls**, which is built on
+Telegram's fork of webrtc.org (**tg_owt**). Both have to be compiled for the
+Sailfish target; there are no packages for them.
+
+That is the whole reason this is opt-in rather than always-on: a plain
+`mb2 build` must not suddenly require a WebRTC tree.
+
+## What has been built and tested
+
+| | |
+|---|---|
+| Architecture | **aarch64** for the media stack (the app itself also builds on armv7hl) |
+| Builds against | `SailfishOS-4.6.0.13-aarch64` and `SailfishOS-5.0.0.62-aarch64` |
+| Tested on | Xperia 10 III, Sailfish OS 5.1.0.11, from a 5.0.0.62 build |
+| Tested against | official Telegram clients (Android, Desktop), both call directions |
+| Scope | 1:1 calls, voice and video. **No group calls / voice chats.** |
+
+The **app** builds on armv7hl as well — verified, and with `voicecalls` off it is
+byte-for-byte the stock build. **Calls** are aarch64-only in practice: nobody has
+built the media stack for 32-bit ARM, which would mean compiling both tg_owt and
+openh264 for that target. `voip.pri` warns rather than pretending, and
+hard-errors with an actionable message when the tg_owt/tgcalls/openh264 trees are
+missing or unbuilt.
+
+Even with a tg_owt for armv7hl, expect video to be marginal there: both encoding
+(VP8/VP9) and decoding run in software. On a comparable 32-bit device a
+GStreamer-based WebRTC client did manage video, but only at poor quality — so
+voice would be the realistic target on armv7hl. i486 is untested entirely.
+
+## Dependencies
+
+Built from source, inside the SDK build engine, for the aarch64 target:
+
+| Component | Source | Revision used |
+|---|---|---|
+| tg_owt | `github.com/Michal-Szczepaniak/tg_owt` | `3215153` ("Add -fPIC") |
+| tgcalls | `github.com/Michal-Szczepaniak/tgcalls` | `8a5b4eb` ("Fix for sfos") |
+| openh264 | `github.com/cisco/openh264` | v2.6.0 |
+
+Those two forks are the ones used by
+[RooTelegram](https://github.com/RootGPT-YouTube/RooTelegram-SailfishOS) and
+Yottagram; they carry the Sailfish/Halium fixes that upstream tgcalls lacks.
+
+Taken from the target's own packages: openssl, opus, libvpx, ffmpeg
+(avcodec/avformat/avutil/swresample/swscale), libjpeg, pulseaudio, alsa-lib —
+plus their `-devel` packages in the build engine.
+
+**openh264 is different.** Sailfish packages it nowhere (patent reasons) but
+tg_owt hard-requires it, so it has to be built from source *and travel with the
+app*: `voip.pri` installs `libopenh264.so.8` into the binary's rpath directory
+next to `libtdjson`, and `__requires_exclude` in
+`rpm/harbour-fernschreiber.yaml` keeps it out of the RPM's dependencies. Without
+that the package would declare a dependency no Sailfish repository can satisfy —
+it would not install anywhere. Point `OPENH264_ROOT` at your build if it is not
+in the default sibling directory.
+
+**libvpx is deliberately not bundled.** It is a real, packaged Sailfish
+dependency, so the RPM requires it by soname (`libvpx.so.9` for the targets
+above) and a device of the same release line satisfies that. Mind the mismatch
+if the device is newer than your SDK: Sailfish 5.1 carries `libvpx.so.12`, so a
+5.0-built package will not resolve there. The clean fix is a target matching the
+device, not a second libvpx shipped beside the system's.
+
+## Building
+
+Everything below runs **inside the SDK build engine** so that the toolchain and
+sysroot are the target's:
+
+```sh
+sfdk -c target=SailfishOS-4.6.0.13-aarch64 build-shell bash
+```
+
+### 1. tg_owt
+
+```sh
+git clone --recursive https://github.com/Michal-Szczepaniak/tg_owt.git
+cd tg_owt && git checkout 3215153 && git submodule update --init --recursive
+mkdir out && cd out
+cmake .. \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DTG_OWT_PACKAGED_BUILD=ON \
+    -DTG_OWT_BUILD_AUDIO_BACKENDS=ON \
+    -DTG_OWT_USE_X11=OFF \
+    -DTG_OWT_USE_PIPEWIRE=OFF \
+    -DTG_OWT_USE_PROTOBUF=OFF
+make -j$(nproc)
+```
+
+This produces `out/libtg_owt.a` (a static library — it is linked into the app,
+nothing extra is installed on the device). X11 and PipeWire are off because
+neither exists on the device; the audio backends stay on because tgcalls' own
+AudioDeviceModule drives PulseAudio.
+
+### 2. tgcalls
+
+Only a checkout is needed — the sources are compiled as part of Fernschreiber
+(see `voip.pri`), so they follow the app's own flags:
+
+```sh
+git clone https://github.com/Michal-Szczepaniak/tgcalls.git
+cd tgcalls && git checkout 8a5b4eb
+```
+
+### 3. Fernschreiber
+
+```sh
+mb2 -t SailfishOS-4.6.0.13-aarch64 build -- \
+    --with voicecalls \
+    --define "tg_owt_root /path/to/tg_owt" \
+    --define "tgcalls_root /path/to/tgcalls" \
+    --define "openh264_root /path/to/openh264"
+```
+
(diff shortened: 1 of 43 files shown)


JimKnopfIoT:
Rebased onto master (0.19 / TDLib 1.8.67) and taken back to a device. Only the
translation files conflicted — both sides had appended `<message>` blocks at the
same place. The call signalling itself needed nothing: `createCall`,
`acceptCall`, `discardCall`, `sendCallSignalingData` and `callProtocol` carry
the same fields in 1.8.67 as in 1.8.21, checked against the new headers.

Then the device found four things. None of them was a consequence of the
rebase — the call code came through it byte for byte identical — they had been
in here all along and only showed up because this was the first run on a Jolla
phone with SailfishOS 5.2.0.17.

## #577 overlaps with this, so I took it over (`14f7cc8`)

#577 ("Show duration and success of (video)calls") touches exactly the two files
this PR touches for `messageCall`. Its rendering is the better one — mine said
no more than "Call (1:23)" — so it is in here now with Simon Schmeisser as
co-author, and with three things repaired that keep it from working as it
stands:

* `fernschreiberutils.cpp` returns `video ? "videocall" : "call"` **before** any
  of the real logic. Everything below that is dead code, and the two literals
  are untranslated. A leftover `qDebug() << messageContent;` sits above them.
* The same file reads `discard_reason` with `.toString()`, but it is a
  `CallDiscardReason` object — the type sits in its `@type`. The QML half does
  it correctly, the C++ half never could.
* Neither half has a fallback. `callDiscardReasonUpgradeToGroupCall` is in no
  list, so in the JS switch it falls through into `messageUnsupported`; and
  `message.content.discard_reason['@type']` throws when the field is absent,
  which it is while a call is still running.

If you take it this way, #577 can be closed as included. @simonschmeisser — the
wording is yours, I hope that is alright.

## Video could not be switched on during a call (`ba1e7b2`)

Telegram has two ways to start a video call and only one of them worked here.
Choosing "video call" outright was fine. Starting a voice call and turning the
camera on while it rings — what the Android client offers as "share video" —
arrived as a plain call and stayed one: no picture from the peer, none sent,
not even a self-view.

`is_video` is the reason. It is fixed when the call is created and TDLib never
revises it; a camera switched on later is negotiated in-band by tgcalls and
never reaches TDLib. Three places asked `is_video` and so could never be right:
`setIncomingVideoOutput` was only called for a call that began as video, so
incoming frames had no sink at all while `remoteMediaStateUpdated` was
reporting the peer's camera the whole time; the capture was created once, at
the start, so `setVideoEnabled()` bailed out for want of one; and the overlay
took `is_video` as "this is a video call".

There is now a button for it, too. `setVideoEnabled()` was `Q_INVOKABLE` from
the start and QML never called it, so video could not be turned on or off
during a call in either direction.

## Three smaller ones

* **`fcd2ab5`** — the button row divided its width by a fixed three ("Up to
  three buttons are visible at once", said the comment above it) and the video
  button made four. They ran off both edges. The share is taken from how many
  are visible now, and from at least three, so a single "Hang up" does not
  stretch across the screen.
* **`101f533`** — the remote video was cropped to fill the overlay. Turned to
  landscape, a portrait frame came out as a strip across the middle of the
  other person. It is fitted now; bars beside the picture are the smaller loss.
* **`0b3f90d`** — a `VideoOutput` keeps presenting its last frame, and the
  remote one is kept visible for the whole call on purpose (hiding it takes its
  surface away and then no frame is ever presented). So the closing picture of
  the previous call greeted the next one while it was merely ringing. The view
  now shows the video only once a frame of this call has arrived, and the
  renderers are cleared as soon as a call with a new id turns up.

## Hanging up when the app is closed (`295d796`)

Closing the app during a call told nobody: the destructor stops the local
tgcalls instance and that is all, while `discardCall` lives in `hangUp()`,
which only the button calls. Measured on the device — the other side sat in a
call that had simply gone quiet.

Sending the discard is not enough on its own, because `sendRequest` only queues
it and TDLib sends it from its own thread; a process that ends right afterwards
takes it down first. The request carries an `@extra` now, TDLib echoes that
back with its `ok`, and `aboutToQuit` waits for exactly that acknowledgement —
usually a few milliseconds, at most 1.5 s, because an application on its way
out must not hang on a network that is gone. A process that is killed outright
runs no code at all; there Telegram's own timeout remains the only thing that
ends the call.

## Built and checked

Built from this branch's exact commit (`295d796`), on top of the current master,
all four measured today on the same machine:

| build | target | result |
|---|---|---|
| `mb2 build` (stock) | 4.6.0.13 aarch64 | OK, 172 s — 0 call symbols in the binary |
| `mb2 build` (stock) | 4.6.0.13 armv7hl | OK, 159 s — same |
| `mb2 build` (stock) | 5.0.0.62 aarch64 | OK, 177 s — same |
| `mb2 build -- --with voicecalls …` | 4.6.0.13 aarch64 | OK, 342 s — 30437 tgcalls symbols, the policy file and 16 multimedia libraries packaged |

Two SDK generations apart, so nothing here is tied to one release. Checked on the
RPM itself with `rpm -qlp` and `rpm -qRp` rather than on the build's exit code:
none of the bundled libraries appears in the package's requires, so nothing is
demanded that no repository can satisfy. Only one file of this PR reaches a stock
build — `qml/components/CallOverlay.qml`, which travels with the `qml` directory
and is never loaded there, because `voiceCallsAvailable` is false. The two
`libtdjson` entries a stock build does carry are upstream's own.

Device: Jolla phone, SailfishOS 5.2.0.17 aarch64 — voice and video in both
directions, video switched on mid-call from either side, landscape and portrait,
and interop with the official Android client.

PR #586, opened 2026-09-12
Send several pictures, videos or documents as one album

JimKnopfIoT:
Fernschreiber already *draws* albums (`MessagePhotoAlbum.qml`, `MediaAlbumPage.qml`)
but could only ever send one file at a time. This adds the sending side, so an
album sent from Fernschreiber renders the way one received from another client
does.

Three commits. The first one stands on its own and fixes albums that are
*received* as well; the other two belong together - see "How the commits
relate" below.

## 0. Half of all albums were thrown away by a sign

`media_album_id` is a random int64 and comes out negative about as often as
positive, but `setMessagesAlbum` admitted an album only for `albumId > 0`. Two
lines further down the same value is tested as `!= 0`, which is what this makes
of it.

Every album with a negative ID was dropped before a single entry was booked, so
its messages kept no members - `MessagePhotoAlbum` draws nothing below two of
them and the album entry filter never merged them into one bubble. **This has
nothing to do with sending**: a received album is dropped by the same line,
which is why albums have always worked in roughly half the cases.

Measured on the device, both albums of one test came out negative:

    albumId -5442158602069634058 type "messagePhoto"
    albumId -6292373871860177921 type "messagePhoto"

and not one call to `updateAlbumMessages` followed. After the change both are
drawn as albums.

## 1. Send several files as one album

**`TDLibWrapper`**

* `sendAlbumMessage(chatId, contentType, filePaths, message, replyToMessageId)`
  sends `sendMessageAlbum`. Telegram takes at most ten items per album, so a
  longer selection goes out as several albums and only the first of them
  carries the caption.
* `newInputMessageContent(contentType, filePath, caption)` builds the
  `inputMessagePhoto` / `inputMessageVideo` / `inputMessageDocument` object.
  `sendPhotoMessage`, `sendVideoMessage` and `sendDocumentMessage` now use it
  instead of each building the same three-field object; the JSON they produce
  is unchanged.

**`ChatPage.qml`**

* The three attachment buttons open `MultiImagePickerDialog`,
  `MultiVideoPickerDialog` and `MultiFilePickerDialog` instead of their
  single-file pages.
* `attachmentPreviewRow.fileProperties` (one file) became `attachedFiles` (a
  list). The preview shows up to three thumbnails and the label carries the
  count.
* A single file still goes out through the existing `sendPhotoMessage` and its
  siblings - an album of one is still an album, and the timeline draws it
  differently.
* The upload progress bar follows every file of an album and the row stays up
  until none of them is uploading. Before, whichever file reported last drove
  the bar and the first one finishing took the row away.
* Picking an attachment now clears the previous one. Choosing a picture and
  then a document used to leave both flags set, and the message went out twice.

**Translations**: one new string, `%Ln files`, filled in for all twelve
languages.

## 2. Keep an album together once its messages have been sent

An album is held together by the message IDs in `albumMessageMap`, and sending
replaces the local ID with the one the server hands back. The map was never
rekeyed - which the TODO at `chatmodel.cpp:690` said outright, and nothing had
ever reached it because albums could only be received.

Alongside it, the guard in `setMessagesAlbum`: `type != "messagePhoto" || type
!= "messageVideo"` is true for every type, so documents entered the album
bookkeeping too. They are never *drawn* as an album - `MessageListViewItem`
loads that component for photos and videos only - but they did get the filter
that hides every entry but the first.

## How the commits relate

The first commit is independent - take it on its own if you prefer, it repairs
received albums either way, and it is the one that made albums appear at all.

The second cannot ship without the first: measured on the device, sending an
album while `albumId > 0` still stood produced three separate bubbles, and part
of the time nothing but empty ones.

The third is reasoned from the code, and I would rather say plainly how far the
measurement goes: `albumMessageMap` is keyed by message ID,
`handleMessageSendSucceeded` replaces that ID, and without rekeying, the map
keeps IDs nothing answers to - which is what the TODO it removes anticipated.
With it in place a freshly sent album is drawn grouped immediately, no restart
needed. What was not done is the counter-test - a build carrying the first two
commits but not this one - so its effect is argued rather than isolated. Drop it
if you see it differently; the other two stand without it.

The document half of its guard is a separate matter: without multi-select nobody
could produce a document album locally, and with it, the first file would have
swallowed the rest.

## What this does not do

Photos and videos cannot be mixed in one album, although Telegram allows it:
Sailfish's gallery pickers are split by type, and each button keeps the type it
always had. Mixing would need the content picker, which is the file browser
rather than the gallery.

## Tested

Built against SailfishOS-5.0.0.62-aarch64 and run on a device with SailfishOS
5.2.0.17 (aarch64): several pictures picked, sent, and drawn as one album in
Fernschreiber right away - and shown grouped in Telegram Desktop as well. Albums
already in the history group on the next read, and a single picture still goes
out as a plain message.


Wunderfitz:
Thanks a lot for this, it's a very thorough PR, and the fix for negative album IDs alone is worth it!

I've pushed a merge commit to your branch to resolve the conflicts with the TDLib 1.8.67 update. Since TDLib 1.8.64, the file has to be wrapped in `inputPhoto`/`inputVideo`/`inputDocument`. `newInputMessageContent()` now builds it with `newInputFile()`, which handles that and also covers the single photo/video/document sends. Please pull before continuing. The merged branch builds for aarch64 against SailfishOS 5.1.0.11.

While reviewing, Claude found two things I'd like to address before merging:

**1. The reply to `sendMessageAlbum` is handled as chat history.**
`sendMessageAlbum` returns a `messages` object. `TDLibReceiver::processMessages()` forwards that to `ChatModel::handleMessagesReceived()` as if it were a history response. All messages are already in the model from `updateNewMessage`, so nothing is added. It still emits `messagesReceived()`, though, and `ChatPage` then calls `scrollToIndex()` for the last read message, resets `loading` and restarts its timers. I'd expect a scroll jump right after sending an album, but I haven't checked that on a device. A cheap fix would be to send the request with an `@extra` (e.g. `"sendMessageAlbum"`) and ignore responses carrying it in `processMessages()`.

**2. Flicker while the album messages are confirmed.**
In `handleMessageSendSucceeded()`, `dataChanged` for the replaced message is emitted before the album is rekeyed. The new `MessageData` starts with `albumEntryFilter = false` and empty `albumMessageIds`. So for a moment the filtered members become visible and the first message has no album members, until `setMessagesAlbum()` restores both. That creates and destroys delegates for every sent message and might explain the empty bubbles you saw at times. Rekeying first, or carrying over the old message's album state, before emitting `dataChanged` should avoid it.

Could you or your agent have a look at these two? Thanks again!

monich:
looks like a claude vs claude fight 🙂

Wunderfitz:
> looks like a claude vs claude fight 🙂

What an epic battle and what a time to be alive...

Wunderfitz:
I asked Claude to implement the open asks and it even found a third thing. Now everything implemented and I'll merge it now. Thank you very much for adding this feature!

**1. `sendMessageAlbum` replies handled as chat history** (`ea9fe0c`)

Went with the `@extra` approach: the request carries `"@extra": "sendMessageAlbum"` and `TDLibReceiver::processMessages()` returns early on it. `processFoundChatMessages()` is untouched, so search results still flow through as before.

**2. Flicker while the album messages are confirmed** (`ea9fe0c`)

Did both halves, because they cover different readers. The album state is carried over from the message being replaced, so `diff()` reports no album role changes and no delegate is torn down. And `albumMessageMap` is rekeyed in place *before* `dataChanged`, because `getMessagesForAlbum()` reads that map directly — QML calls it synchronously from `onAlbumMessageIdsChanged`, so it has to be right at the moment anything is emitted.

**3. Albums of eleven** (`d423a22`)

`sendMessageAlbum` takes 2–10 messages. Chunking in steps of ten left a single item over for every eleventh file, so eleven pictures went out as an album of ten and an album of one. A chunk now gives up one item when that keeps the next from standing alone — eleven go out as nine and two. Caption placement is unaffected; it still lands on the very first item only.

Issue #587, opened 2026-09-16
Sent voice messages appear as zero length in all other platforms


cosmonad:
Hi Sebastian,

I've been following Fernschreiber development lately as I've just received a Jolla Phone 2026, this client is amazing and so back to the basics, thank you so much for all your efforts!

I'm encountering the following issues as of today:

- Recording and sending a voice message causes it to appear as if it were 00:00 in length on the other end (both iOS 26 and Android 16, latest Telegram versions): the audio actually plays back, something related to metadata I guess?
- Something similar happens the other way around: if I receive and listen to a voice message it is not marked as such in their clients.

Both bugs could be somewhat related. I've compiled your latest changes hoping something along the way fixed them but unfortunately didn't. I'll see if I can figure out the root issue unless you've already found it :)

Have a great day,
Eugenio

cosmonad:
Got it working, opened a pull request!

Wunderfitz:
Forgot to close, thanks for the contribution.

PR #588, opened 2026-09-16
Fix voice note duration, waveform and listened state

cosmonad:
Well, gave Claude a spin to see what Opus was actually capable of doing and, oh well, I'm impressed. Here's a fix for https://github.com/Wunderfitz/harbour-fernschreiber/issues/587 I've reported myself a couple hours ago. Kept the madlad under strict control and added comments myself where necessary, but it did a good job.

I've tested the fixes on my Jolla Phone 2026 against all platforms I have access to and worked out as expected on all of them: iOS 26, Android 16, Web K, Web A and Desktop on Linux. Missing a Windows machine and a Mac, but I'm confident it should work there as well.

So, recapping what the pull consists of:

- The outgoing voice notes appeared with no duration nor a waveform on every other client: those attributes are now sent along the message itself. The waveform is built by collecting amplitude peaks during recording using QAudioProbe (looked promising from the docs). If for whatever reason the probe can't attach to the recorder, we simply log the event and ignore the waveform, leaving the voice note unaffected (apart from appearing as a flat line on the other end).

- Listening to an incoming voice note never marked the message itself as read: turns out viewMessages marks a message as read, but only openMessageContent marks its content as opened, which is what sets is_listened, and that request was missing entirely.

Hoping this will be helpful, whish you a great evening.

Cheers,
Eugenio


cosmonad:
Minor update here, fixed a tiny issue.

The openMessageContent response only confirms the request, while the actual state change arrives separately through updateMessageContentOpened: since that update was previously unhandled, cached messages kept their old is_listened/is_viewed state unless fetched again.

I've added one last paragraph to the commit message as I may have found a bug in TDLib: a voice note in Saved Messages stays unlistened regardless. I was able to replicate using Telegram Web but not Telegram Desktop which I guess speaks MTProto directly... Could really be a TDLib issue.

I'll see if they already have something related in their issues.

Wunderfitz:
Thank you very much for that fix. Happy to merge it!

